PR #2, opened 2021-11-30
charmsy

Matheusu:
I added the possibility for the charms to have sockets, the small one and only one, the medium one and two, and the big one and three.


CollinHerber:
We will likely not be adding this mod. Perhaps some version of it but it is far too powerful in the current state of the other mod.

CollinHerber:
Feel free to post this on the #user-files of discord however.

PR #5, opened 2021-12-31
Fixed copy & paste error with Diamond and Emerald Gem Bag credit recipes

yourb3stnightmare:


CollinHerber:
This PR adds the STACKED version of these to the cube receipies. Which could cause you cubing a stack of 10 and losing the entire stack.

yourb3stnightmare:
This is not the pull request you think it is.

This pull request updates the input of two existing recipes that had the wrong input gems. The diamond is supposed to be a standard diamond (gsw), but it was set to a perfect diamond (gpw). The emerald is supposed to be a standard emerald (gsg), but it was set to a standard diamond (gsw).

PR #27, opened 2024-05-13
controller ui rework

ND-State:
Woof figuring it out. Controller UI rework to be similar to PC ui


CollinHerber:
Got to play with it this morning, looks great. Thank you so much!

PR #51, opened 2024-09-25
Update runes.txt

Malvar01:
corrected not working skills in runes.txt on many runewords


CollinHerber:
Validated it ran. Not sure what was all changed here to validate however.

PR #58, opened 2024-09-30
Fixed Lightning Sentry skill text to reflect that Death Sentry was removed as a synergy

Bakelt:


CollinHerber (commented):
Looks good to me but there is a merge conflicts. Might have taken the file from master and not next. Might need to branch from next instead.

Issue #75, opened 2024-10-04
Peril hasn't the intended stats (it seems)


ympadilha:
What I got:

![peril-bug](https://github.com/user-attachments/assets/d1011a31-219b-4895-8bcb-284e9897d7f2)

What it should be:

![peril-what-should-be](https://github.com/user-attachments/assets/9ba629dc-5826-475a-90f1-b9d18b0050f9)

PR #76, opened 2024-10-04
fix: attributes for Peril runeword

ympadilha:
# Proposition to change the Peril runeword

I opened an [issue](https://github.com/D2R-Reimagined/d2r-reimagined-mod/issues/75) regarding Peril's attributes being incorrect, I did some testing and realized that min and max damage are limited to 63-127 respectively, given that the weapon's damage was not nearly enough the initially proposed,

I'm proposing replacing:

```
dmg-min		100	200
dmg-max	300	400
```

With:
```
dmg%		250	250
dmg/lvl	        25
```

I copied those 2 attrs from Windforce.

Here's how it looks like now (level 66 while looking at the item):

peril-proposition


CollinHerber:
There was a whitespace issue causing merge conflicts. Not sure what editor did that. 
Here I added your changes: 5ef4bae275c868b10d5f94ab78e25909fdcf9f9b

I realize that means ur not marked as a contributor however I will still accept you as if you did as a Mod Contributor. Reach out in discord for your role.

PR #77, opened 2024-10-05
Fix Necromancer Skeleton PetMax

svceon:
The original formula is min(lvl/2, 14).
My proposed formula is min(1+(lvl-1)/2,14).

The original one has a slight flaw that does not work like the notes:
"Maximum of summoned skeleton warriors is +1 per 2 skills"

to work this way my formula works better, here's a comparison between the two formulas (A = new, B = original):

image


Maowiwowi:
Probably apply that to the Druid wolves as well, they have that same pattern for the additional wolves.

CollinHerber (approved):
LGTM! Thank you for the contribution. Reach out to me in Discord to get your role.

PR #78, opened 2024-10-05
Maowiwowi- Essence of socketing/conversion Part 1

Maowiwowi:
These 3 parts change the Essence of Suffering(Essence of Socketing) into essentially a jewelers orb from PoE. They also change the Essence of Hate(Essence of Conversion) into an Orb of Chance for magic items but always be a unique and keep the base type.

I haven't adjusted any drop tables as I'm not sure the rate that would be balanced. Im guessing the Socketing should probably drop a bit less frequently than jewels currently are. At the moment we're able to willy-nilly socket any decent base we find, I think there should be a bit more friction there.

Essence of Socketing:
It's random sockets so they might get 1 or max in a single tap. The current socketing is more consistent but it removes the enjoyment of finding the right base with the right sockets. Having some variation and potential upset is always worth the feeling of a 1 tap and brings back a bit of that excitement.

Essence of Conversion:
With the number of uniques in the game, I thought a way to target farm would be a great addition. Especially for the items that could potentially be build defining. It's not limited in what items it can work with as long as there is a magic base to use it on and a unique version to upgrade to.


Maowiwowi:
Sorry for 3 submits back to back. I cannot figure out how to make changes to 3 files in 1 pull request. Messed with this stuff last night, didn't change any drop rates or anything like that. Additionally, I left the Essence of Conversion to upgrade only Magic Items to give you a reason to potentially want to check for magic bases for unique item target farming.

CollinHerber (approved):

PR #80, opened 2024-10-05
Maowiwowi- Essence of socketing/conversion Part 3

Maowiwowi:


data/local/lng/strings/item-names.json
```diff
@@ -50540,4 +50540,4 @@
         "ruRU": "Perfect ÿc4*ÿc8Skullÿc4*ÿc0 Stack",
         "zhCN": "Perfect ÿc4*ÿc8Skullÿc4*ÿc0 Stack"
     }
-]
\ No newline at end of file
+]
```
CollinHerber, line 50543: Same comment as before, instead of changing existing. Lets create net-new item

CollinHerber (changes requested):

PR #92, opened 2024-10-06
Heartcarver - Replace Grim Ward with Deadly Treasures

Bakelt:
Grim Ward is a dead skill. Replaced with Deadly Treasures since it replaced GW on the skill tree.


CollinHerber (approved):
LGTM!

PR #96, opened 2024-10-06
Re-corrected Light Sentry Tooltip

Bakelt:
Removed Death Sentry from synergies on tooltip as the skill has no synergy after Blizz 2.4 patch.

I believe this change got blown out at some point.


CollinHerber (approved):
LGTM

PR #98, opened 2024-10-06
Update treasureclassex.txt

Maowiwowi:
Adjusted drop rates for Orbs


Maowiwowi:
I think you'll need to roll back the push I made directly to next before this that adjusted the table already to resolve the conflict.

Issue #106, opened 2024-10-07
Name issue for Diamond Mail unique


CollinHerber:
![image](https://github.com/user-attachments/assets/d69c1e86-6189-483d-a753-389ac30adec6)

Gotta find the unique this matches to and then debug the name.
Likely either duplicate ID or missing entry in item-names.json

Nick-Conrad:
Hey "Celestial Revelation" is also having the same issue. I was going to fix it on my branch but since you're looking at it.
Act 1 Rogues are having the issue with their definition of "Rogue" lol

![image](https://github.com/user-attachments/assets/3b1310f1-5c56-4dc6-adaf-b836616dedb1)


![image](https://github.com/user-attachments/assets/dc956b5d-edf2-4463-b8d7-35a3ee0e0259)

PR #115, opened 2024-10-07
fix: adamantine links unique item name showing an evil force

CollinHerber:
Resolves: #106 

Fixes an item showing An Evil Force


Nick-Conrad (approved):

Bakelt (approved):
ye

PR #116, opened 2024-10-07
fix: holy fire numbers were not matching the vanilla

CollinHerber:
Fixes the holy fire damage and synergies not matching vanilla


Nick-Conrad (approved):

Bakelt (approved):
burn baby burn

PR #117, opened 2024-10-07
fix: change irongolem petmax to 1 to prevent save issues

CollinHerber:
Change the petmax to 1 for iron golem to prevent deletion on save


Bakelt (approved):
Only one can sit upon the iron throne

PR #126, opened 2024-10-08
cubemain.txt - Categoraziation and minor changes

Harald92:
### CHANGES:
- Categoraziation of all recipes for easier overview and handling. The next step would be to reorder all recipes in an alphabetical order. 
- Streamlined recipe format
- Deactivated the old socket punch system recipes
- Changes the ethereal item tranform recipes. Now the Item will keep its stats instead of being rerolled, usefull for making superior base items ethereal without losing the superior buff.

### ADDITIONS:
- Added the new socket punch system recipes
- Added the new jewel upgrade recipes


Bakelt (changes requested):
This reverts the T5 to T6 splash fix. Need that merged in.

Also, T4/5/6/7/8 should have their input3/4 changed from "gemy" to "gem4" to use unstacked gems like the rest. See below. Better for you to just merge it.
image

PR #127, opened 2024-10-08
Duplicate Skill Issue#121

### LINKED ISSUE #121: Skills.txt has duplicate "proc_SplashDamage_Four" skill
skilldesc is unique so it might not cause any actual issues
image

Bakelt:
Duplicate proc_SplashDamage_Four should be Five

skilldesc is OK so I think this is non-functional change

Resolves #121


CollinHerber (approved):
LGTM

PR #128, opened 2024-10-08
Remove splash visual

Bakelt:
Had complaints of lag. Lets remove the visual until we have a solution that lets it be optional.


victoria-riley-barnett:
2nding this, I didn't have lag personally but the visual was way overdone and I couldn't aim at mobs (which is still required with this splash radius tbh)

PR #129, opened 2024-10-08
Holy Bolt Synergy Reduction

Bakelt:
Reduced holy bolt synergies from 50% to 40%. Net loss of ~20% damage at full synergy.

Note: Also contains the same resolution for #121


data/global/Excel/skills.txt
```diff
@@ -100,7 +100,7 @@ Smite	97	pal	smite		150
...
+Holy Bolt	101	pal	holy bolt		8										holybolt			holybolt																																																																												paladin_holybolt_cast											cast_undead				17						holybolt	holybolt											1	7		none												SC	SC	xx												1	1							1							paladin_holybolt_cast	cast_undead		6	25														1	1					1	4	32	1	1	1										1 + stat('extra_holybolt'.accr)	Projectile Count											1	HP Heal Min baseline	2	HP Heal Min per level	6	HP Heal Max baseline	4	HP Heal Max per level					15	Healing synergy	40	Damage synergy									1								8															mag	8	8	10	13	16	20	16	8	11	15	18	23	(skill('Blessed Hammer'.blvl)+skill('Fist of the Heavens'.blvl))*par8								384	3000	0
```
CollinHerber, line 103: This also needs skilldesc update doesn't it?

data/global/Excel/skills.txt
```diff
@@ -376,4 +376,4 @@ proc_SplashDamage_Three	373		proc_SplashDamage_Three	4	8													proc_splash
...
 Mark of the Wolf	376		mark of the wolf		18																			markwolf		4500		item_tohit_percent	par1	item_maxhp_percent	par2																																																																																																	0		none																		1																1				1						1									2	wolf					0	0																																30	Attack Rating %	40	Life %																					1								8																																				256	1000	0
-proc_SplashDamage_Four	377		proc_SplashDamage_Five	4	8													proc_splashdamage_five																																																																																										11	17						proc_splashdamage_five												1	7		rng		miss	miss									A1	A1	xx				1																1										1															1						1	8	4	1	1	1										min(24,ln12)	# missiles	par3	activation frame	2	triggered							11	number of arrows to fire	1	additional arrows per level	1	Acivation frame of arrows.																			1								8	128																																			384	3000	0
```
CollinHerber, line 379: Looks like this may have reverted?

Bakelt: Looks ok on my end? Four changed to Five.
There is a separate pull req for just this change but figured I'd merge it here if I was doing a separate update that impacted skills.txt

data/global/Excel/skills.txt
Bakelt: Modification is for 'Holy Bolt'par8 which is used for both the synergy damage calc in skills.txt and the tooltip in skilldesc.txt
Tooltip is auto updated in this case since the formula didn't change, just the constant.

CollinHerber (approved):
LGTM

Issue #134, opened 2024-10-08
Runeword "Starlight" - Less MF than stated MIN on website


RegularAlicorn:
## Steps to reproduce
* Make starlight 

## Issue
Starlight has 180% MF

## Expected Result
Follwing the website, the Runeword has **230 - 380 MF**, the minimum should be 230 not 180 (or false information on the website)

CollinHerber:
We'll want to update the docgen to figure out why this is an issue and resolve it. Moving issue to Docgen repo.

PR #135, opened 2024-10-08
changes starlight MF values matching websites data

RegularAlicorn:
see issue https://github.com/D2R-Reimagined/d2r-reimagined-mod/issues/134


CollinHerber:
I think instead of adjusting this we should fix the docgen so the website accurately reflects. This is already an insane amount of mf for a single item and boosting it more is probably not the best idea.

PR #139, opened 2024-10-08
Wake of inferno tooltip synergy fix

Bakelt:
Changed Fire Burst from par7 to par8 (same as WoF)
Removed Death Sentry from tooltip
No impact to actual damage. This is just a tooltip correction. Verified synergies against D2R


CollinHerber (approved):
LGTM

PR #149, opened 2024-10-09
fix: duriel and andy dropping nothing in tz

CollinHerber:
* Fixes Duriel Drop Tables in TZs.. Hopefully finally
* Fixes Andariel Drop Tables in TZs.. Hopefully finally


Bakelt (approved):
Invest in TP scrolls. Their value is about to rise

PR #151, opened 2024-10-09
fix: cowking dropping insane amount of corruption orbs

CollinHerber:
* Reduces number of corruption orbs dropped from cowking


Bakelt (approved):
Moo moo moo, moo moo. Moo.

PR #153, opened 2024-10-10
Amazon updates

CollinHerber:
* Make it so ice arrow pierces by default
* Makes it so amazon skills don't use quantity from offhand when using skill


Bakelt (approved):
Bringing her on up from F-Tier to E-Tier

CollinHerber:
> Bringing her on up from F-Tier to E-Tier

Pretty much xD

Issue #156, opened 2024-10-10
Add loot beam to Orb on Conversion


CollinHerber:
With how rare this item is, I think it's worth adding a loot beam to it.
Will need to create a json file if not already exists. There are examples in the project already

victoria-riley-barnett:
Is this still relevant or can we close? I see 
<img width="722" alt="Screenshot 2024-10-22 at 10 39 26 AM" src="https://github.com/user-attachments/assets/352ce614-1253-4759-bbc3-0316aff926cb">

victoria-riley-barnett:
But havent dropped a conversion in 1.21

CollinHerber:
This is completed. Just not closed.

PR #158, opened 2024-10-10
#156 / Missing Synergy for Double Throw

Bakelt:
Added Frenzy synergy. This was mentioned as a new synergy in Character Balancing menu


CollinHerber:
> Added Frenzy synergy. This was mentioned as a new synergy in Character Balancing menu

I've not played double throw but I feel like people said it was good already, does it need more dmg?
Just wondering if it would be better to remove the char balance text instead. Unsure which is the best

Bakelt:
I’ll do a quick respec on my barb this evening and see what the damage looks like in comparison to frenzy. I think I have a good throw stick of two around

CollinHerber (approved):

PR #161, opened 2024-10-11
Updated set naming in files

Harald92:
- Renamed the sets and set items so their names are displayed like all other sets and set items. Now they are represented with their actual name to make editing easier.


CollinHerber (approved):
LGTM! Nice

PR #162, opened 2024-10-11
Unique changes

Harald92:
- Set "nolimit" for all uniques to 1 to enable uniques appearing multiple times in the same session to make it harder to min/max for specific uniques with the new Orb of Conversion or for example gambling.
- Set "lvl" to an appropriate value in relation to "lvl req" to reduce the drop chance of high level items by the item not dropping from low level enemys. This was somewhat regulated by the base items _lvl_ inside _armor.txt_ and _weapon.txt_ but with this change it is made sure.


CollinHerber (approved):
A bit hard to read through this one, but I think it's alright.

PR #165, opened 2024-10-11
fix: fix being unable to stack one out of stack of 100

CollinHerber:


Bakelt (approved):
Took one down, passed it around, 99 Rune of Zod on the stack

PR #166, opened 2024-10-12
Fix/Change: Item name changes & fix

Harald92:
- Changed name of potions to MP/HP 1-5 for easier identification of potion quality
- Fixed Amethyst stack coloring
- Changed Rune 20-33 coloring format. Rune name is orange like all other runes now, pick up text is still in red.


Harald92:
I don't know why this PR shows so many changes, I compared this file vs the one of the next branch and the changes I made are correct and there are only a few lines changed.

CollinHerber (approved):
I dig it

PR #169, opened 2024-10-12
feat: add beam to conversion orb

CollinHerber:
Adds light beam to conversion orb


Harald92:
Do we get those for each orb? I am a very big fan of the pillar for rare and important items because loot text can get desrupted pretty easily and having to listen to item drop sounds to indicate good loot is kind of stupid.

CollinHerber:
> Do we get those for each orb? I am a very big fan of the pillar for rare and important items because loot text can get desrupted pretty easily and having to listen to item drop sounds to indicate good loot is kind of stupid.

Corruption orbs drop from bosses that specifically won't be surrounded by tons of drops so I think that's okay. They are also pretty common. Conversion are super rare and can drop anywhere so I think just conversion for now.

Bakelt (approved):
Looks good. 
image

PR #173, opened 2024-10-12
Fixed Sting RW for Amazon

Doxorn:
Sting RW didn't give +5 Ice Arrow if equipped as Amazon so I gave +2 Ice Arrow for Amazon Only
Sting BEFORE
Sting AFTER


Doxorn:
For amazon only Sting only gave +3 to Ice Arrow. For everyone else it gave +5 as stated.

Bakelt (commented):
That is an oskill (+skill with no class tag) and they are intentionally hard coded in the engine to not give the full value to the class the owns the skill. They are hard capped to 3.

Doxorn:
That I didn't know. I guess this isn't a bug then.

Harald92:
> That is an oskill (+skill with no class tag) and they are intentionally hard coded in the engine to not give the full value to the class the owns the skill. They are hard capped to 3.

Is there a reason for it? Do oskill give synergys or anything like that?

Bakelt:
No synergy bonus. You'd have to ask the guy from 20 years ago, but my guess is that they wanted to be able to design items that gave a large amount of +oskill so that other classes could attempt to make fun builds with the skill from another. For example, Flamebellow gives +12-18 to inferno. Getting +12-18 to a skill off one item on top of your other +skill and skill point investment would be absolutely imbalanced. However, it allows other classes to work with a decent skill level as a base. 
image

Harald92:
> No synergy bonus. You'd have to ask the guy from 20 years ago, but my guess is that they wanted to be able to design items that gave a large amount of +oskill so that other classes could attempt to make fun builds with the skill from another. For example, Flamebellow gives +12-18 to inferno. Getting +12-18 to a skill off one item on top of your other +skill and skill point investment would be absolutely imbalanced. However, it allows other classes to work with a decent skill level as a base. image

Makes sense.
I noticed that other classes can use a skill without themself having the extra dmg of synergies said skill would need to be a very high level, granting that would overpower the class which got synergies for the skill.
I have another solution to that tough, creating the same skill just for that item which can be carried by other classes. :D

Issue #105, opened 2024-10-06
Clickable Area of Cube Incorrect


CollinHerber:
![image](https://github.com/user-attachments/assets/3b7b62ba-e56b-4628-a27e-8bd7d9673dd2)

The right side after the transmute button should not click through and send the move command like it is now

CollinHerber:
Both stash and inventory

Orin55:
![CUBE](https://github.com/user-attachments/assets/e0f6a102-aaab-4ee2-8809-b9b894357b79)
![STASH](https://github.com/user-attachments/assets/58c3bbdf-d8fb-44e8-909d-94a7795d97dd)

These areas outside of the red box to be precise.
These "clickable" areas in the stash and cube screens are misaligned with the new UI.
If you click outside of the red marked area, it gives a walk command to your character, closing the stash screen.

PR #174, opened 2024-10-13
FIX for issue #105 : Clickable Area of Cube Incorrect

Doxorn:
Areas where the sidebars supposed to be still not clickthrough, but at least areas that should not be are not.


CollinHerber (approved):
Nice, this looks good to me

PR #178, opened 2024-10-13
qol: cubemain.txt Update 2.0

Harald92:
- Unified crafting recipe name format
- Added input & output to all recipes names
- Ordered recipes name in alphabetical order
- Disabled item stamina upgrades
- Disabled item durability upgrades
- Disabled specific Item lvl upgrade recipes (I want to introduce universal reroll recipes for uniques/set items later)

No Recipes were changed, only their name and ordering inside the file.


CollinHerber (approved):
It's a bit hard to judge all whats going on but I trust ya

Issue #183, opened 2024-10-14
Duriel Loot Table Broken for Non-TZ


Bakelt:
Man somehow dropped less loot than vanilla D2

PR #184, opened 2024-10-14
#183 Fixed non-TZ Duriel Loot Table References

Bakelt:
monstats was referencing a deleted "Duriel" instead of "Duriel - Base" for each of it's non-TZ variants

Verified on a normal non-TZ non-quest kill
image


CollinHerber (approved):
LGTM, ty

PR #185, opened 2024-10-14
Improvement: revamp item skills

Harald92:
- Replaced all kill-skill, levelup-skill, death-skill, skill, oskill, gethit-skill, hit-skill, att-skill and charged skill Ids with the actual skill name.

Used ChatGPT for that one. Had to fix some stuff by hand in the end and overviewed the changes, they look good.


Bakelt:
Tests needed:
1) Force a unique drop and verify it has the updated and working skill
2) Verify existing unique doesn't brick 

Note: I believe the others reference skill by "skilldesc" and not "skill" column. I don't know if there is a functional difference as the only differences I see are capitalization and skills unavailable to players not having a "skilldesc"

Harald92:
> Tests needed:
> 
> 1. Force a unique drop and verify it has the updated and working skill
> 2. Verify existing unique doesn't brick
> 
> Note: I believe the others reference skill by "skilldesc" and not "skill" column. I don't know if there is a functional difference as the only differences I see are capitalization and skills unavailable to players not having a "skilldesc"

If you do not beat me to it I will test mutliple uniques tomorrow. I want to test every possible skill value at least for 3 items, so it would be 27 Items to test.

Harald92:
Tested all changed prop values and the skills are displayed correctly and are working. 
Used some uniques I already have and they are skill working too, the neat part is that all already dropped weapons keep their stats from before a change. I noticed that behavior when testing new items, for a change to take place you have to always get a new unique of the same kind to see the changes in item attributes.
Skill prop Check

Doxorn (changes requested):
.json file must be updated for the changes in set item index names and set names (otherwise you get "an evil force" string)
Line 796 "Chaoskill er" corrected to "Chaoskiller" since "er" were in a new column, shifting every cell to the right by 1. With the fix it doesn't cause crash on startup
an evil force

Harald92:
> .json file must be updated for the changes in set item index names and set names (otherwise you get "an evil force" string) Line 796 "Chaoskill er" corrected to "Chaoskiller" since "er" were in a new column, shifting every cell to the right by 1. With the fix it doesn't cause crash on startup an evil force

Did the change for "Chaoskill er" to "Chaoskiller" resolve the name or does it still presist? I would think it should be resolved with this becaue formating is correct again.

Harald92:
> > > .json file must be updated for the changes in set item index names and set names (otherwise you get "an evil force" string) Line 796 "Chaoskill er" corrected to "Chaoskiller" since "er" were in a new column, shifting every cell to the right by 1. With the fix it doesn't cause crash on startup an evil force
> > 
> > 
> > Did the change for "Chaoskill er" to "Chaoskiller" resolve the name or does it still presist? I would think it should be resolved with this becaue formating is correct again.
> 
> The "Chaoskiller" item ingame have the "Chaos Flight" name since it wasn't updated in the **..\data\local\lng\strings\item-names.json** item-names.json chaoskiller_ingame

There are multiple Items which have a different name ingame than in the files. I would make that problem a new issue rather than fixing that with this PR.

Doxorn (approved):

PR #187, opened 2024-10-14
Created some Key Conversion recipes

Nick-Conrad:
Not sure if we want this, have seen many people say it'd be nice. This gives an alternative to forced farming but I'm not sure if the cost is high enough, this will possibly loose purpose to really target farm summoner and nithlahak if you're rolling in gems. Idk up to you.

Terror -> Hate = Terror + 2 P Diamonds
Hate -> Destruction = Hate + 2 P Skulls
Destruction -> Terror = Destruction + 2 P Emeralds


Bakelt:
I wouldn't hate having a recipe but it does feel cheap. Nihlahak will be much easier to farm next patch as the tomb viper bug should be fixed.

Harald92:
Could be a chance to give essence a use.

CollinHerber:
Saving this for 2.0 as well

CollinHerber:
I think I'm okay with this recipe for now, but we should revisit it

CollinHerber (approved):

PR #196, opened 2024-10-17
Druid SS Duration Buffs

Bakelt:
- Removed duration on werewolf and werebear. They are now permanent and can be removed by casting the spell again
- Lycanthropy no longer gives duration to werewolf or werebear. It now gives duration to maul and feral rage buffs at +2seconds/lvl


CollinHerber (approved):
LGTM

PR #199, opened 2024-10-18
Rat Lords Curse, Blanched Death string fixes

### LINKED ISSUE #177: Missing Strings
Missing strings for:
Serpent Lord
Blanched Death
Rat Lord's Curse
Medusa's Gaze(?) - Someone had two but only 1 was missing a name and had a different sprite.

victoria-riley-barnett:
fixes #177 (atleast in part)

### DIFF
--- a/data/local/lng/strings/item-names.json
+++ b/data/local/lng/strings/item-names.json
@@ -30465,7 +30465,7 @@
     },
     {
         "id": 50262,
-        "Key": "Rat Lord’s Gate",
+        "Key": "Rat Lord’s Curse",
         "enUS": "Rat Lord's Curse",
         "zhTW": "Rat Lord's Curse",
         "deDE": "Rat Lord's Curse",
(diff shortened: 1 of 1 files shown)


Bakelt (commented):
Blanched Death look's good. 

I'm a little confused on why Rat Lord's Curse was broken as it looks like it does exist as "Rat Lord's Gate" in uniqueitems.txt which matches the key. I don't know if the report is old or if this was "fixed" to match the key at another point. 

I don't know if we want to change uniqueitems.txt to also be "Rat Lord's Curse" or if we want to revert the .json change and assume the person who reported the issue is on an old patch and it has since been fixed.

CollinHerber (approved):
Seems fine, BakeIt's mention is weird but yeah if it works it twerks

victoria-riley-barnett:
I'll change the name in uniques.txt too, it's likely someone updated everything except the key because its uniqueitems.txt key was still "Gate".

PR #200, opened 2024-10-18
removed magic arrow/bolt (zo1,zo2) spawns from akara/fara, add aqv,cqv

### LINKED ISSUE #171: Reminder: Remove Magic Arrows and Magic Bolts from Akara
These bolts aren't the same as charsi bolts and won't work with crafts

victoria-riley-barnett:
Confirmed working with successful arrow and bolt crafts from akara. Separate issue did occur where piercing recipe only had replenish quant, but I think fix is in latest cubemain.

Fixes #171


CollinHerber (approved):
LGTM, ty!

PR #201, opened 2024-10-18
Rat Lords Curse

victoria-riley-barnett:
Updates uniqueitems.txt key to match item-names.json key


CollinHerber (approved):
Idk yo, seems sus

PR #203, opened 2024-10-19
Bank/Stash moved into view for more resolutions (e.g., Steam Deck)

rb1t:
Prior to this change, the left side of the Bank/Stash was cut off on my Steam Deck (something I enjoy playing D2R**R** on :smile: ). E.g., the left-most Bank/Stash slot is not visible on screen, and you can't interact with it (e.g., pick up a rune from the slot off-screen).

From my testing, this works well for the Steam Deck, so hopefully it will help other devices as well. I also tested this on two larger resolutions in hopes there wasn't a major regression. Specifically, the resolutions I rested with:

- 3840x2160 
- 2560x1440
- 1280x800 (the native resolution for the Steam Deck).

All  seem to work fine.

If I'm doing this wrong or misunderstanding, I'm happy to help test/resolve.

--- 

 _EDIT/NOTE_: 
 
 - At first, I thought  it wasn't scaling properly. But after overlaying two screen shots (stretching 2560x1440 up to 3840x2160) I can see the position looks to be scaling fine.
 - However, testing on the more extreme / narrow resolution of 1280x960, I can once again see the left being cut off. Changing the x value to `1535` here makes it very functional for controllers, but perhaps at the cost of aesthetics (i.e., no gap between the bank and player items)._


rb1t:
Recreated as #214, sorry for the noise.

PR #208, opened 2024-10-19
Remove cubemain levelreq mod1 for base upgrade recipes

### LINKED ISSUE #207: Upgrade basetype of set weapon can brick usability
image

1h sword bricked in both known cases (my own, and pictured from Tetra)

victoria-riley-barnett:
Resolves #207 

Basetypes update their level using the recipe just fine without additional cube mod. This adds "hidden" req. level (shows on character, but a level 58 expecting to wear their elite weapon cannot equip it to tell it's now req.level 60), with strange behavior depending on if user is wearing another set piece or not (see #207)


Harald92 (approved):
Looks good.

CollinHerber:
I think this was intentional from the 2021 days but I think it's alright to remove it

PR #216, opened 2024-10-21
added item-names.json sorting/renumbering script

victoria-riley-barnett:
- script confirmed works, I ran it and dropped the sorted and renumbered item-names.json into my game. 
- Would you believe we actually only have ~4000 entries 😆 

Usage: node item-repair.js

### DIFF
--- a/scripts/item-repair.js
+++ b/scripts/item-repair.js
@@ -1,0 +1,34 @@
+const fs = require('fs');
+const path = require('path');
+
+// Define the file path
+const filePath = path.join(__dirname, '../data/local/lng/strings/item-names.json');
+
+// Read the JSON file
+fs.readFile(filePath, 'utf8', (err, data) => {
+    if (err) {
+        console.error('Error reading file:', err);
+        return;
+    }
+
+    // Parse the JSON data
+    let items = JSON.parse(data);
+
+    // Sort the array by the `id` field
+    items.sort((a, b) => a.id - b.id);
+
+    // Reassign the `id` field to be sequential starting from 1060
+    let newId = 1060;
+    items.forEach(item => {
+        item.id = newId++;
+    });
+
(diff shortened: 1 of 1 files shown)


victoria-riley-barnett:
One note on running this: some other open PRs touch this file, they should all be accepted before we run it, I think

Harald92:
Just out of curiosity, what does that script do? It sorts the item-names.json that much I understand but in what fassion?

victoria-riley-barnett:
The IDs are completely arbitrary, and the way they're setup now its pretty likely someone will reuse an ID eventually and not realize it. This just makes it so that never happens, it sorts everything by ID, then renumbers them starting from 1060 (the first ID rn, probably arbitrary but I left that lol)

victoria-riley-barnett:
I guess it could still happen, but it cleans up the odds of that happening and gives an easy fix: someone just has to run this on item-names.json, and then commit the new file.

Harald92:
> I guess it could still happen, but it cleans up the odds of that happening and gives an easy fix: someone just has to run this on item-names.json, and then commit the new file.

Very nice and very usefull indeed. You can commit it too, or do you not want to hastle with that? I have zero background in script stuff and are a bit hesitant to run a script I do not 100% trust on my computer.

victoria-riley-barnett:
> One note on running this: some other open PRs touch this file, they should all be accepted before we run it, I think

I was concerned abt running it now with PRs that touch item-names.json in the pipeline, I don't mind being the one to run it, but also anyone familiar with JS in the future will see this doesn't do anything bad lol

victoria-riley-barnett:
You'd need node for it anyways

victoria-riley-barnett:
hmm, going to close this for a bit until I figure out whats up - on a 2nd test with 1.21 fully setup, changing this file in this way caused CTD.

Issue #211, opened 2024-10-19
Myhrginoc's Deathmonger gives conflicting mastery


Bakelt:
Gives +10 to Shield Mastery (previously Axe Mastery ID)
Set uses a 2h axe so this does nothing for the set. Recommended to change to +10 to Melee Mastery

CollinHerber:
Resolved via #219

PR #219, opened 2024-10-21
FIX Myhrginoc's Deathmonger

Harald92:
Fix for issue #211

- swaped the set items perk of Axe Mastery to Sword Mastery so it gives Melee Mastery ingame instead of Shield Mastery


CollinHerber (approved):
LGTM. As long as we swapped it in skills, because pretty sure it was still Sword Mastery. Assuming it was updated and I just missed it. I remember reading about it.

PR #225, opened 2024-10-21
slight buff for unique jewel hunting to make up for abundance of poor…

victoria-riley-barnett:
… craft outcomes now

- In my testing on 1.21 (accelerated jewel crafting), the current rarity of desired unique jewels feels very high in comparison to how long they take to get. Instead of adjusting that, this adds another outlet of use for "poor" uni jewel outcomes to chase the good elemental/level/40/15/15 facets.


Bakelt (approved):
Seems fine to me.

PR #229, opened 2024-10-21
Removed * from second SoJ

Bakelt:
There is no * in the item-names.json and I doubt we should have special characters in the uniqueitems.txt
image


victoria-riley-barnett:
Tried this locally - item remains missing string fwiw

PR #247, opened 2024-10-23
1.21.1 custom menu update

Doxorn:
### Custom menu:
- Added Strafe exception to Amazon changes list
- Removed Duriel from bugs
- Added Not A Bag section

### Not A Bug section:
- oskills +3 limit explained
- paladins take damage when using necromancy


CollinHerber (changes requested):
Needs to be branched from `patch` not `next`

PR #260, opened 2024-10-24
Add Charm Weight System

CollinHerber:


CollinHerber:
automagic doesn't appear to be applying to all weapons as intended. Appears to be pointing to +1 attack. Gotta solve this.
Also any weapon not pointing to 312 needs to be updated to the cloned automagic group.

Didn't have time to resolve but posting here for reminder/clarity. Anyone is welcome to check it out

Bakelt:
Looks ok. skills.json IDs are invalid but we already have plans to revise those in #265

Bakelt (changes requested):
Change Required:
skills.txt
- Hidden Charm Passive should be moved to the end and index updated. Currently it will offset all the new skills which breaks monprop.txt

Questions/Comments:
automagic.txt
- Whats the point of the Fletcher's2 and so on? Shouldn't the affix just be appended to existing charms?

misc.txt
- Whats the point of the elixir change?

weapons.txt
- what is the point of adding the new automod groups 312 and 316 to the weapons? Why aren't they consistently 312 or 316?

CollinHerber:
> Change Required: skills.txt
> 
> * Hidden Charm Passive should be moved to the end and index updated. Currently it will offset all the new skills which breaks monprop.txt
> 
> Questions/Comments: automagic.txt
> 
> * Whats the point of the Fletcher's2 and so on? Shouldn't the affix just be appended to existing charms?
> 
> misc.txt
> 
> * Whats the point of the elixir change?
> 
> weapons.txt
> 
> * what is the point of adding the new automod groups 312 and 316 to the weapons? Why aren't they consistently 312 or 316?

> * Whats the point of the Fletcher's2 and so on? Shouldn't the affix just be appended to existing charms?

You can only have a single auto prefix on an item, so those are ones that are specifically designed for items that already had auto prefixes to keep them.

>  Whats the point of the elixir change?

Didn't do anything here. This appears to be some random thing that I never touched. I have zero context to this.

> what is the point of adding the new automod groups 312 and 316 to the weapons? Why aren't they consistently 312 or 316?

Not sure what you mean by this.

Delegus:
Okay so initial pass to get everything working properly.

Automagic updated the names to reflect what they do because they don't reference a string anyways. Necro heads require all 3 fields to add the poison damage to physical attacks so changed to poison skill and poison pierce instead to accommodate weight system and because the added poison damage to physical was pointless.. Added oSkill to necro heads and paladin shields.

Armor.txt Added weight oSkill to all shields. (Applies to magic/rare/crafted) Shields that required level 1 have automagic (auto prefix column) 316 applied which reduces level requirement to 1 instead of 7 because of hardcoded level increases. All other shields have automagic (auto prefix column) 312.

itemstatcost modified levelreq again so it can roll -6 to 249  to accommodate low level item level requirements. Lowered charm weight priority so it's always the bottom affix. Raised oSkill priority so it is only second to "Corrupted" in mods list, this will affect all oSkills listings on items, but I would rather see them at the top anyways.

Misc.txt added auto prefix column so that charms will always spawn with the charm weight amount from automagic. Removed elixir changes as these are column headings and appear to be unintentional. It is not spawnable anyways.

Setitems.txt Added weight oSkill to all weapons and shields.

Skilldesc.txt Added string to missing columns in case the game ever references them.

Skills.txt Changed passivecalc1 to 41 so up to 40 will work and mana is set to 0 at 41 or after. Removed erroneous multiplayer functions.

Uniqueitems.txt charms updated to include the charm weight property. All weapons and shields updated to include weight oSkill.

item-modifiers updated so + or - required level is displayed properly since it can now go up to -6 to accommodate level 1 item drops.

Weapons.txt I only changed a column that was causing diablo clone to spawn when selling hand axes. Other columns are set to respective automagic entries based upon whether the item is a level 1 or if it is higher, and to accommodate weight system and still spawn staff mods for staves orbs, or class skills for amazon specific weapons.

Delegus:
item-modifiers.json
- Fixed string error so attributes are all orange.

Bakelt (approved):
LGTM - maintained Fire Golem rework changes

Delegus (approved):
Or can I? Yep, can review, so this is finally done.

PR #263, opened 2024-10-25
Club/Mace/Hammers now share all runewords

Bakelt:
- This has only caused confusion as blizzard labels them all "mace class" even though runewords are very particular
- Intentionally left out scepters as they can roll +class skills and may have unintended power creep with things like HotO


CollinHerber (approved):
LGTM

PR #271, opened 2024-10-28
Ele Druid Rework

Bakelt:
Elemental druid skill synergies previously forced the player to spend the vast majority of their points in the elemental tree and only allowed them to take "1 point wonders" in other trees like Oak Sage. By reducing the reliance on synergy bonuses, an elemental druid should now feel more free to pick up summons or even elemental skills from the other tree.

- All elemental skills now have a synergy bonus soft cap of 25 points and an uncapped bonus past 25 points at a reduced rate
- Synergy bonus was adjusted so that the bulk of the previous synergies (25-75 points) are now in the first 25 points. 
- Synergies were changed from the previous limitations to now work with ALL elemental skills (e.g. Fissure can gain more fire damage from Tornado)


Bakelt:
I think I'm done with updates on this for a bit and it is good for merge whenever. Tested a 1-95 fresh and 95-99 with access to the mules. Seems to be in a good state.

CollinHerber (approved):

Issue #275, opened 2024-10-29
Sunder Charms


CollinHerber:
Adding in Sunder Charms to the Drops in Hell Terror Zones

This also means needing to make the sunder charms themselves as they have not been added yet.

Check the v2 branch for implementation

PR #286, opened 2024-11-02
Sunders

CollinHerber:
Resolves: https://github.com/D2R-Reimagined/d2r-reimagined-mod/issues/275
Adds in the sunder charms and also changes the savepath to Reimagined


Delegus (approved):
LGTM; Resistance itemstat changes break compatibility; Need to advise on new shortcut and save path on future play tests and final release.

PR #294, opened 2024-11-05
Class Rework - Singer Barb

Bakelt:
Howl
- Has been reworked into a short range cone effect that does physical damage.
- Synergy bonuses from all shout based skills

War Cry
- Now deals a small echoing shockwave from each target hit for a percentage of the skills damage
- Skill curve has been redone
- Synergy bonus extended to include the other shout based skills

Battle Cry
- No longer reduces enemy defense
- Now reduces enemy physical damage reduction

Shout
- Duration synergy extended to include all shout based skills

Battle Orders
- Duration synergy extended to include all shout based skills

Battle Command
- Duration synergy extended to include all shout based skills


CollinHerber (approved):
Looks fun

PR #295, opened 2024-11-06
Match lesser potions sprites

VtorHdev:
The lesser_potions sprite looked out of place, too bright. Adjusted brightness and contrast values ​​to match the rest of the potions.

Actual
image

Modified (darker)
modified_sprites


Bakelt (approved):
I like it

PR #304, opened 2024-11-16
Orb Stacking

Vingelis:
Allow corruption, socket and conversion orbs to stack


CollinHerber (commented):
I don't think we actually need a separate item for the "Stack". The reason we needed seperate ones for runes and gems were because they can be inserted into gear, that is not the case here.

I think it should just be a matter of making it stackable, and then updating the recipies that exist that use it to make sure it's a single one and then add new recipes to split a stack.

Vingelis:
> I think it should just be a matter of making it stackable, and then updating the recipies that exist that use it to make sure it's a single one and then add new recipes to split a stack.

If players need to start from scratch in 2.0 then this approach is doable, otherwise it will break save files as it changes existing items in players' inventory/stash.

CollinHerber:
> > I think it should just be a matter of making it stackable, and then updating the recipies that exist that use it to make sure it's a single one and then add new recipes to split a stack.
> 
> If players need to start from scratch in 2.0 then this approach is doable, otherwise it will break save files as it changes existing items in players' inventory/stash.

We are planning on releasing other breaking changes as well, hence the major version bump so it's all good.

Vingelis:
Have been trying to make this work but doesn't seem to play nice.

1) Item description text can either be what's defined in mosc.txt (spelldescstr), or quantity if no spelldescstr is defined - you can't have both.

2) Stacking/unstackcing orbs works fine, but if you cube an orb by itself you can no longer stack it with others until you restart the game.

3) Corruptions don't work, even when updating the cube recipes to take 'ka3,qty=1' as its input, for example.

CollinHerber:
If it comes down to needing to do the whole cube nonsense to stack and unstack like gems and runes are, then I'd rather not even make them stackable tbh.

CollinHerber:
My guess is something is just not right.

Issue #310, opened 2024-11-18
Investigate Sound Issues


Bakelt:
Some sound settings aren't impacted by the sound sliders. 

e.g. Sin Kick, hit sounds

PR #311, opened 2024-11-18
Resolves: #310

Bakelt:
Appended new sounds to vanilla D2R sound file and corrected duplicate index at start of new files
Resolves: #310


CollinHerber:
Can't review the changes cause too big lol. Verified the custom sounds still work?

Bakelt:
> Can't review the changes cause too big lol. Verified the custom sounds still work?
Not sure what the sound effects are tied to. I dropped a few rune stacks and they still made sound.

Bakelt:
Updated out of date merge

CollinHerber:
So wait a sec, you just copied in vanilla and appended our extras? We have over 200 custom sounds , and there are some edited rows as well. So if that's the case they were unintentionally reverted

Bakelt:
> So wait a sec, you just copied in vanilla and appended our extras? We have over 200 custom sounds , and there are some edited rows as well. So if that's the case they were unintentionally reverted

Correct. ID 11391 to 11469 were re-appended to the vanilla file. Any changes to the existing vanilla rows would be blown out. We can either merge those back or look into a proper fix.

Bakelt:
Did another quick test. Only replacing the "Channel" column with vanilla data seems to correct the issue. Will need to investigate what is broken there.

Bakelt:
It looks like we have blank channels for some sounds while Blizz file has none. Intention was probably to remove these sounds, but it looks like that might brick the feature. Blizz file has blank Redirects but no blank Channels or blank FileNames

CollinHerber (approved):
Resolves need to go in the commit body, added it for you. Should now close automatically.

PR #314, opened 2024-11-19
QoL-CraftingArrowBolt

Bakelt:
Reduced ilvl of crafted arrows and bolts to lvl1 to allow for crafting at any level. Should reduce the amount of complaints of failed arrow crafts.

Required level remains untouched.


CollinHerber (approved):
Will reduce confusion, although I think we should revisit how these are obtained in the future

PR #316, opened 2024-11-21
Cube Recipe Additions

Vingelis:
Recycle Recipes for Set/Unique Jewelry.

Cubing 3 unique or set rings / amulets + ID Tome creates a rare jewel.


Bakelt:
Seems like a reasonable idea to me. Unsure of what ratio would be good as you currently see a LOT more unique/set rings/amulets than rare jewels.

Vingelis:
I based this off of my own experience so far. The rate that I find unique rings is more frequent than amulets, but still about 3:1 ratio of set to unique jewelry.

Vingelis:
I don't mind amending to 3U/3S -> 1 Rare Jewel if others agree that that ratio makes sense. Personally I think 3 is about right.

CollinHerber:
> I don't mind amending to 3U/3S -> 1 Rare Jewel if others agree that that ratio makes sense. Personally I think 3 is about right.

I'm fine with 3 as well

Vingelis:
Updated to require qty=3

Bakelt (approved):
LGTM

PR #326, opened 2024-11-24
Rune Stack Changes

Vingelis:
1) Create Rune Stack: Cube a single rune to convert it to a stack with quantity of 1.

2) Stack Runes: Cube a single rune and a rune stack of the same type to combine them. Rune stack quantity increases by 1.

3) Unstack Runes: Cube a rune stack to create a single, useable rune of the same type. Rune stack quantity decreases by 1.

4) Upgrade Runes: Cube a rune stack (quantity of 2 or more) and a TP Tome to create a single, useable rune from the next tier up. Rune Stack quantity decreases by 2. Can still cube two single runes of the same type to create a rune from the next tier up.

5) Downgrade Runes: Cube a single rune, or a rune stack, with an ID Tome to create a single, useable rune from the previous tier. Rune Stack quantity decreases by 1.


Delegus:
Testing breaks any stack with "useitem" as a return and causes the magic property "item-qty" to stack in either + or - until you S&E and rejoin the game; Further testing so far I haven't found a fix as I don't know of a way to read/modify quantities with the cube aside from the misc stack options at the bottom of cubemain.txt; As is, this isn't workable.

Delegus (changes requested):
Changes discussed in detail; Cannot use item-qty magic property for any recipes as it causes major stacking problems prior to save and exit for any stacks involved in a recipe applying this property; Stack/Unstack and all recipes should be changed to not include modification of item-qty property.

Delegus:
Closing as unsolvable currently; Reopen or resubmit if a solution is found.

PR #327, opened 2024-11-25
Cube Reroll Rare Amulet

Delegus:
3 ammy = ammy currently when the recipe states 3 ammy should = ring


Delegus:
Need to split these

PR #332, opened 2024-11-26
Fixes Magic Prefix

Delegus:
Prefix name had a 1 in front causing missing string error.


Delegus:
Closing this because I'm doing further investigation as to why there are so many prefix/suffix with a # after it; If anyone can clarify that would be awesome.

PR #342, opened 2024-12-01
Skill Hotkeys

Delegus:
1.21.1 - 2.0 Preview causing blz-log.text error "Cannot find JSON reference ($SkillHotkeyTextAltStyle)" so investigated and the entire reference for the skill row icon bar thing was missing so D2 was tossing default there and it wasn't compatible with the old _profilehd.json which used multiple workaorunds to at least show some kind of key. I got it working at 1080P and doesn't alter controller layout/behavior; Could someone test with a higher resolution monitor for me?

Edit: Screenshots on my most hot keyed character; Have tested on others using spacebar (Space) numpad keys (Show up as Num0-9) and shift and looks fine. Dropshadow changed to no transparency but can dial it back to default 200 of 255 if it is too glaring.

image
image

### DIFF
--- a/data/global/ui/layouts/_profilehd.json
+++ b/data/global/ui/layouts/_profilehd.json
@@ -429,24 +429,34 @@
     "SkillHotkeyTextStyle": {
         "fontColor": "$FontColorLightGold",
         "dropShadow": {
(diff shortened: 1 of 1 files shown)


CollinHerber (approved):
LGTM

PR #346, opened 2024-12-04
Fix Merc Ring Slot

Delegus:
Added the second ring slot to mercs so they will stop taking a second ring without being able to remove/adjust it. Looked into disabling the slot entirely and was unable to find/figure out how, so I just added it into the hireling inventory and updated the sprites.

Fixes both controller and non-controller game views.


CollinHerber (approved):
LGTM

PR #350, opened 2024-12-04
Update Add Skills to Amulets

### LINKED ISSUE #343: Recipe Conflict
6P Skull + Rare = Rare Reroll conflicts with only amulets 6P Skull + Rare Amulet = +1 Necromancer;

Need to change the vanilla recipe or change all 7 add stats recipe to require 1 less (already too powerful IMO) or 1+ more additional P gems.

Delegus:
Resolves #343

Adding skills is already powerful, instead of changing vanilla recipe to reroll, updated all 7 class skills to require 9 P Gem instead of 6


Vingelis:
Can we not change the recipe to require unique jewels as well? Anything that provides +Skills should be expensive. This is part of the crafting requirement for adding +Skills when crafting pieces and requires about the same rate of materials for the output.

Delegus:
This is an update to resolve a conflict; Balance isn't the goal here. But it can be updated to include that if people weigh in with an opnion.

Bakelt:
> This is an update to resolve a conflict; Balance isn't the goal here. But it can be updated to include that if people weigh in with an opnion.

I agree. Make a separate issue to capture that we want to revisit this for balance but get the bug fix in now

Bakelt (approved):

Issue #345, opened 2024-12-03
Terror Zone Improvements


Bakelt:
Recommended that we redo the rotating TZs to allow more choices at once. You can have a MAX of 20 desecrated levels at a time. Expanding the choices would allow for better QoL for summoners.

Each rotation should include multiple options. Probably 1 act boss, 1 key boss, and at least 1 high density area with a few other thrown in. Variable terrain options would also be ideal (indoor / outdoor).

Bakelt:
Act by act works fairly nice outside A4 but the zones there might be big enough to not matter. Maybe we can push new zones there if we make any new portal based ones.

<https://docs.google.com/spreadsheets/d/1gQrxKpopQvrjSfB0uhe--1meYthH1SNN55DFW0v5bdI/edit?usp=sharing>

PR #352, opened 2024-12-04
TZ Revamp

Bakelt:
Resolves: #345

Rotates between 5 TZ sets which are Act1-5 (cows in A4 due to size limit).


Delegus (approved):
I ran around for a few hours off and on while working and it cycled and worked fine on my 1.21.1 sorc; Never got any crashes or missing zones. Only change from then to now is the level increase.

Issue #355, opened 2024-12-05
Uber Duriel and Mercs


Delegus:
Uber Duriel sometimes doesn't draw mercenary aggro for quite a while, requiring you to kill mobs around him and let him beat on you before the merc will attack.

PR #360, opened 2024-12-05
Fix Duriel Threat

Bakelt:
Resolves: #355 

Set uberduriel's threat parameter to 14 like the other ubers

Already tested this fix during NM Hell rebalance where pets would ignore regular Duriel for everything else due to empty threat parameter


Delegus (approved):
Quick and clean

Issue #337, opened 2024-11-28
Set Orbs


Delegus:
We have socketing orbs, corruption orbs and conversion orbs. 

Currently there is no orb to make an item into a set.

Proposed: Cube conversion alone = Set orb. Cube set orb alone = Conversion orb.

Idea: Since we added recycling into rare jewels from unique/set jewelry, rare jewel*9 + 1 unique or +1 set = Orb of that type.

Make conversion set orbs work the same way as the finalized idea in issue 336 based on either A: required level B: required level +8 or C: random chance based on rarity.

Currently rare jewels are just a stopgap between magic and rare aside from the few people that use them to socket set items they luck into. This will make it easier to get set items you wish, give rare jewels more of a purpose, and take away from unique item jewel chase using recycle recipe.

Vingelis:
Have addressed the initial proposed idea and left the other ideas concerning recycling to be raised in a new issue.

PR #362, opened 2024-12-06
Orb of Assemblage

Vingelis:
Resolves: #337 

New orb that allows for conversion to set items.


CollinHerber (changes requested):
I think it's rather mundane to just have a recipe that converts them back and forth. I'd instead make a treasureclass entry that picks one of the two to spawn

Vingelis:
Have removed the cube recipe to convert the two and instead added the new orb to the Jewelry TC as suggested.

Delegus:
This is going to cause a string issue; With how things are currently coded you need an ooa for the name and an ooaDescription for what the item does based on current text files.

Delegus:
This is going to cause a string issue; With how things are currently coded you need an ooa for the name and an ooaDescription for what the item does based on current text files.

CollinHerber (approved):
LGTM

Issue #300, opened 2024-11-06
Investigate Min & Max Damage Property Scaling


CollinHerber:
![image](https://github.com/user-attachments/assets/9505e219-4394-417a-a2f0-fa0293a213ec)

Some of our items have higher minimum and maximum damage than what the property allows.
Since we are doing save breaking changes anyway, lets see if we can bump min & max damage as well so it doesn't "break" into 63-127" damage

Issue #160, opened 2024-10-10
Too much MF from 1 Source overflows


Bakelt:
Similar issue happens with Deadly Treasures barb passive skill.
![image](https://github.com/user-attachments/assets/8db8b720-202f-47d4-bf41-19ee100061be)
![image](https://github.com/user-attachments/assets/6bc99e13-c1a8-4ac6-a1d5-3301e6889857)
![image](https://github.com/user-attachments/assets/a5624fa8-6182-4dba-8ba9-8c81a5513e65)

Weirdly enough, it looks like even though it is signed, there is some funky behind the scenes stuff to prevent negative MF.

wjakiela:
I confirm. I have the same in Deadly Treasures case. If I add point to Deadly Treasures and overload some mf level it doesn't add more mf. Despite that I can add mf from other sources.

Harald92:
Did you test if making 2 MF prop would work?

Bakelt:
For DT, this could be a possible solution. Split it into 2 and just display the sum of the two.
For items, too much of a pain and Collin has mentioned a lot of the items from Talonrage used something else to increase the caps. We may not even want that powercreep and the integer limit is just a bug acting as a balancing tool.

Harald92:
Not more of a pain as reodering the cubemain file. Powercreep is an important bullet to dodge, but balancing stuff is never easy.

Harald92:
The adding item stat system via cube has the same problem, opened another Issue for it _**Stat overflow for Item upgrading via Cube #179**_

Issue #179, opened 2024-10-13
Stat overflow for Item upgrading via Cube


Harald92:
The recipes for adding item stats has many problems with overflow of item stats. 
The whole system is kind of OP in the long run and the recipes cost is setup kind of weird too.
If there is no solution to the overflow problem disabling the recipes should be considered.

Delegus:
#585

PR #367, opened 2024-12-08
Itemstatcost and strings

Delegus:
- Resolved: #300
- Resolved: #160

- Related: #179
-Items will always have a maximum stat limit based on Save Bits in itemstatcost.txt and soft coding can increase the limit but will approach item disappearing if the values are too large.

- To Block (Increased % Chance to Block) Updated per Unique Aegis Lachdonnan's Guard from 0-63 to 0-127 as the prior cap was 63%
-image

- Min/Max/Throw Flat updated from 63-127 to 255-511 due to various items, but notably Adamantine Bow Hydra Bow
image

- Flat Magic Find % Updated from -100 to 155 to -100 to 411 due to Starlight rune word and Heaven's Treasure Archon Plate
image
image

- Item Level Requirement updated from 127 to 255 to require many more - level requirement recipes to bring the item to useable after spamming add item stats.
-String added as a low priority to display at the bottom of item tooltips.
image

- Replenish quant/dur fix altered for future compatibility; Flat 1024 gold cost increase for the affix for buy/sell because prior affix was maxing out the buy/sell when paired with any skill affix on the same item.

- Collin's Charm updated from -50 to 1,125,899,906,842,573 to -100 to  411 due to apparent typo in save bits and save param bits
-Updated strings to clarify x% melee splash
image

- Gembag string updated to reflect new maximum added in the beginning of 2.0 of 65,535 credits
image


Bakelt (approved):
LGTM

PR #369, opened 2024-12-08
Necro Rework - Blood Golem, Blood Mana, Bone Skills

Bakelt:
### Teeth, Bone Spear, and Bone Spirit
- Damage curves have been increased
- Energy synergy has been added
- Magic Mastery has been added

### Golems
- Blood Golem now provides a Blood Mana aura to the player causing them to use health instead of mana
- Blood Golem now provides Magic Mastery
- Golem Mastery now provides added bonus damage similar to Skeleton Mastery
- Stat synergies between golems have been removed and instead baked into the golems themselves. (e.g. no longer need to invest into Clay Golem for attack rating)

### Weaken
- "reduced physical damage" has been reduced
- Now also reduces enemy magic resistance

### Monsters
- Regurgitator series HP has been reduced for Nightmare


Delegus (commented):
Removing the blood mana curse from multiple enemies will significantly increase the power of sorceress players who only stat into energy and rely on energy shield for mitigation; Is this intended?

Bakelt:
> Removing the blood mana curse from multiple enemies will significantly increase the power of sorceress players who only stat into energy and rely on energy shield for mitigation; Is this intended?

There is no way to have both as Blood Mana is a hardcoded state. I'd rather just nerf the power of energy shield later

Delegus (approved):
Curse change was my only concern, looks good to me.

PR #370, opened 2024-12-09
Update Corruptions in Cubemain

### LINKED ISSUE #278: Corruption Orb Updates
* Make more rare across all difficulties
* Add more downsides. Such as downgrading item to white.

Delegus:
Resolves #278 

- Socket change outcomes have been completely removed as between Orb of Socketing and Jewel punch these recipes were just frustrating and pointless.

- Overview:
-- All items now have a 10% chance to turn white (or be rerolled as magic for amulets and rings) and consume the orb
-- All items now have a 10% chance to be rerolled as rare and consume the orb
-- All items now have a 25% chance to do nothing and consume the orb
-- All items now have a 55% chance to give a beneficial stat of some kind with the upper limits remaining as untouched as possible.
-- Other low to middle values have been slightly adjusted to fit into the new % scope.

- All items can now be clicked twice while the second click will consume the orb regardless of outcome.

- Any item that has already been clicked twice will now return the item and the orb while still making a transmutation noise so the player can be fully confident the corruption process has been completed successfully.

- Additions:
-- Belts can now roll 10% faster cast speed with a 1% chance as the 100 roll was missing from the file. 

- Tested using a modified misc.txt file to sell corruption orbs for 1 gold on Akara and gambling random magic/rare items from Gheed, each type was rolled 100 times and the outcomes were consistent with values listed.


Bakelt (approved):
LGTM - saw some test results in discord

PR #371, opened 2024-12-09
Orb Sprites

Delegus:
Updated all 4 orbs to a new sprite custom to D2RR

image

Orb of Assemblage description color updated to be consistent with other orbs.

image
image


Bakelt:
LGTM - I think we may want to eventually change the string color from all red to the color that matches the orb.

Bakelt (approved):

Issue #204, opened 2024-10-19
Missing String for a magic affix


Bakelt:
I believe the new repair+ethereal is the culprit
![image](https://github.com/user-attachments/assets/143b904f-1144-416b-a7f4-07310bce6644)

victoria-riley-barnett:
It's name (Ghostly1) as well as several other ones is shared with the eth/rep-dur/att% mod - could that be the issue?

victoria-riley-barnett:
Also unsure if these names need to be referenced elsewhere, data guide not clear on that it makes it sound independent

Delegus:
Many of these have numbers (which can be safely just be removed as long as they reference a valid stat) or reference an invalid stat entirely that doesn't exist in properties or itemstatcost; ex attribute % (so I marked them as not spawnable instead of removing and messing up the rows) I have been playing (leveled a paladin 1 to 99) since making a rough change to my own files but haven't updated a Git branch for it yet in case someone knows more than just me investigating; Maybe a holdover of the mod merge that wasn't included in the string json files.

PR #372, opened 2024-12-09
Magic Prefix and Suffix String Updates

Delegus:
Resolves: #204

- magicprefixes.txt and magicsuffixes.txt compared against item-nameaffixes.json in order to attempt to update all missing strings
-- Disabled suffixes Agility, Power, Endurance, and Spirit, as they reference completely missing stats.

- Added ENUS to all strings that did not have a populated string in translations.

- Any conflicts will be resolved and changes re-uploaded as need, but GitHub isn't showing any at the moment.


Bakelt (approved):
LGTM - downloaded and verified some string updates

PR #374, opened 2024-12-09
Key and Key Grabber Sprites

Delegus:
Updates Key and Key Grabber Sprites to be visually different from each other.
image
image
image
image
image
image


CollinHerber (approved):
LGTM

PR #384, opened 2024-12-17
Zon Bow Rework

Bakelt:
## Amazon
- Missile range for magic/fire/cold arrow and multishot set to same values. 
- Magic Arrow now gains some flat magic damage per skill point
- Magic, Cold, and Fire arrow now have +1 Projectile per 3 levels of Multiple Shot.
- Magic, Cold, and Fire arrow now have reduced base physical damage to match Multiple Shot
- Penetrate now provides -6% to Enemy Ele Resists and -2% per hard point

## Druid
- Spirit of Barbs Ele pen updated to match Amazon's Penetrate


Bakelt:
Ready for review. Will need to revisit Ice Arrow when we come up with an idea.

Delegus (approved):
Looks good in game, nice work.

Issue #344, opened 2024-12-03
Reroll Recipes - Stack > 3 P Gem + Magic Item


Delegus:
When a stack is modified with item-qty magic property it holds that - # and applies it to any future items added to the stack. Example, use a stack of 100 gems to roll 3 items, that property is now -9 so now you make a stack of 3 of that gem and drop it in, it gains +3 - 9 overall quantity; This property is remedied after Saving and Exiting before dropping new stacks onto the existing and already recipe'ed stack - Same issue as the rune pull request #326 

Recipes affected - 
![image](https://github.com/user-attachments/assets/9f7bda20-8f5e-449d-abd7-58dc75c8689b)

It says any item but it's only the 3 different charm bases listed as inputs.

PR #386, opened 2024-12-18
Bugged Cube Recipe

Delegus:
Resolves: #344 

Convenience does not override bug or broken. Current iteration of this recipe allows for infinite + or - gem stacking when used with multiple stacks. Recipe removed until a better solution can be found.


Bakelt (approved):
LGTM

Issue #336, opened 2024-11-28
Unique Item Drop Level


Delegus:
Unique items are inconsonant between vanilla and mod added uniques. All vanilla uniques are on a curve from +3 to +8 to +3 level vs required level, where there is a wildly inconsonant level to required level for all mod added unique items.

Proposed changes:
A: Make all unique items drop level = required level
B: Make all unique items drop level = +8 of required level
C: Only change conversion orbs to always drop an ilvl 100 item so that treasureclass and rarity determine chance of getting a specific unique

A requires copying column J to column I in uniqueitems.txt
B requires copying collumn J to column I with +8 using a created script or manual labor in uniqueitems.txt
C: Requires changing cubemain.txt u1058 to blank and t1058 to 99, u1077 to blank and t1077 to 99; u1079 to blank and t1079 to 99; u1080 to blank and t1080 to 99; u1088 to blank and t1088 to 99; 

I propose change C simply because, the recipe can be used on any level char to roll any unique of the base input and therefore eliminates wasting conversion orbs to get a rare item return; Gives a "I have this 81 required item and I'm 73, need to play more"; Takes out level component and bases the item rolled entirely on RNG from rarity value.

Alternatively, I propose option A as a secondary consideration as it prevents people from using a mule to reroll but doesn't punish them for not knowing you need to be 8 levels higher than the required item level.

Finally C: Vanilla unique roll behaviour, however we would need to apply the 3 to 8 to 3 curve to all mod added items to make the drop levels consistent with required levels.

Bakelt:
Do we know what impacts the uniques rarity when there are multiple options? For example, I believe Lustwander are more rare than the other Mirrored Boots option. We probably want to maintain that to some degree.

Delegus:
Yep, that's why I just updated the cube recipe so conversion orbs always roll a 99 item so it will always return a unique of some kind. When considering the various rarity fields, rarity of orb itself and required drop level of Stairway is 66 with 74 equip and rarity 2 while lust is 74 with 81 equip and rarity 1, it's going to work the same for anyone who was at least level 74 rolling for them in the past with the pull request change. 

Overall this issue is likely going to need to be revisited when you start doing hell drop rebalancing and most of the mod unique items are droppable by 80 except any requiring 84 or 85+ to equip

PR #387, opened 2024-12-18
Unique & Set Drop Levels

Delegus:
Resolves: #336 

Equalizes all unique and set item drop levels to be equal to their equippable levels.

Future updates to drop rarity will be easier as the only field that needs to be changed is rarity.

Future updates to required level to determine drop location in difficulties are easier.

Future updates to orbs can be reconsidered based on character level with an accurate and consistent recommendation of character level versus required level.


Bakelt:
Are there more changes pending for this?

Delegus:
Was supposed to be a draft, need to go through and make sure there isn't anything else like Annihilus and Hellfire or any mod added items that just won't drop now.

PR #392, opened 2024-12-19
Hireling adjustments

Bakelt:
### Mercenaries
- Added new Merc Skills for A3 merc based off the original sorc skills
- Increased scaling for Charged Bolt, Lightning, Firebolt
- Adjusted synergies to be based off the skills the merc learns
- A3 merc no longer benefits from +sorc skills, only untagged +all skills and untagged +fire/cold/lightning skills


Delegus (changes requested):
Would like to get charm weight in before this and have the skills.txt here IDs updated so I don't have to go through and change 1000 oSkills.

Delegus (approved):
And let's get this in too, looks good.

PR #396, opened 2024-12-21
Channel Updates

Bakelt:
- Updated some skill sounds to their original blizzard Channel setting


Delegus:
So no matter what I change the multishot sounds to, it still randomly fires off extremely loud. Starting to think it is in the arrow projectiles and not the multishot sound. The other changes work fine.

Delegus:
The bug this was meant to fix has been fixed in a previous PR, we can open a new one if we want to re-address sound channels.

PR #399, opened 2024-12-22
Issue 393: Rune Stack Changes

### LINKED ISSUE #393: Rune Stacking / Unstacking / Upgrading / Downgrading
Current implementation of rune stacking uses 'useitem' modifier to increment / decrement item-qty. When used this way the engine will duplicate the current item's modifiers. When combining stacks generated in this manner the item modifiers (+ or - item-qty) is duplicated depending on the number of times the action was performed on the original stack. This may lead to situations where stack quantities increase or decrease by an undesired amount.

We need to rework the current implementation of stacking to not use useitem in conjunction with item-qty to avoid these unintended side-effects.

Vingelis:
Enhancement to rune stacking, unstacking, upgrading and downgrading

Approach has been revised from original submission, moves away from using useitem in conjunction with item-qty. Now we check for item quantity using cubemain Op functions specifically from 1-100, and output stacks using 'usetype' with quantities from 1-100. This ensures that we never duplicate magic properties that have unintended side-effects.

Persisted upgrade rune recipes using two unstacked runes.
Updated downgrade single rune recipes to also require an ID Tome.

Closes #393


Delegus (approved):
I can't break the system, looks good to me. Had a low spec PC user test and only a second increase in loading time. Cube main theoretical limit is 600K lines, so merging.

Delegus:
In the future please include clear notes of what each change and recipe does so that we don't have to dig through the file changed to interpret what each recipe ends up doing. Updating the 2.0 notes was a pain and I don't know that I got it all. I just know I've tested this since you uploaded it while testing my sorc changes and between here and the Discord kinda knew what was going on.

PR #401, opened 2024-12-22
Assets for Sprites

Delegus:
Requires GIMP or Photoshop to edit. Export as a PNG after manipulating the layers. Will need to edit key borders and export as a png then import them into the grabber file to update.


CollinHerber (commented):
To be consistent with the rest of the directing casing I'd make it `assets`

PR #403, opened 2024-12-23
uniqueitems.txt Fixes

Delegus:
Fixed Missing Property Reference
Fixed oSkill property on 6 items because it was added in the wrong field.

Recreated request because I forgot to do it on a branch.


Bakelt (approved):
LGTM

PR #407, opened 2024-12-24
Clay Golem HP

### LINKED ISSUE #333: Clay Golem exceeds life cap
Player reports clay golem dies in 1 hit. 
During golem rework, we should adjust the formula to either reduce the HP gain and/or make it have diminishing returns.
image

Delegus:
Resolves #333 

Caps out at skill 100 with 25 hard points in golem mastery and clay golem  at 34969 life.
image
image


Delegus:
Sigh skills.txt is broken here also, need to remake the pull request.

PR #408, opened 2024-12-24
Conversion and Assemblage Drop Fix

Delegus:
The class "Conversion Orbs" needs to be above the classes that are trying to nest from it, relocated above Jewelry so these orbs will properly drop.


Vingelis (approved):
LGTM

Vingelis (approved):
LGTM

CollinHerber (approved):
Had no idea this was a thing. That's annoying asf

PR #409, opened 2024-12-24
Necromancer & Tooltip Cleanup

Delegus:
Resolves: #333 

Clay Golem Max Life is now 34969 at skill level 100 and consistent with other golem HP scaling on current rework. Tooltip updated. Par1 reintroduced at a much lower value. Golem tested against mobs in all 5 acts, Diablo and Baal.

image

Amplify damage tooltip now updates to reflect correct physical damage taken value based on level.

image

Cleaned up other tooltips using the same StrSkill37 value as Amplify Damage to be positive instead of negative so they make sense in game.

image
image
image


Bakelt (approved):
LGTM

Issue #410, opened 2024-12-27
Socket Punch - 40th attempt?


Delegus:
We have no socket punch for crafted items.

Do we want unique jewels or rare jewels to apply to crafted items?

Can't find where it was never added intentionally but if that's the case then this can be closed.

PR #412, opened 2024-12-28
Socket Punch Crafted Items

Delegus:
Resolves: #410 

Updated socket punch to include crafted items.
Crafted items will require unique jewels = # of sockets trying to be added.
Base item and ilvl restrictions still apply.


Bakelt (approved):
LGTM

PR #414, opened 2024-12-29
Sorc Missile Fix

Delegus:
When merged in now only initial missile does damage.

This fixes that.

It appears I updated the skills.txt in my game folder and while testing other changes overwrote the update in my sorc update branch since this was the last issue I fixed before readying the pull.


Vingelis (approved):
LGTM

PR #420, opened 2024-12-29
Hell Archer Nerf

Bakelt:
- Reduced the damage bonus for archers in Hell-revamp as they were already deadly


Delegus (approved):
Archers are less deadly per attack, but small hit chance buff.

Issue #265, opened 2024-10-26
.json IDs are capped to 65,536


Bakelt:
We are probably overwriting some texts in various places with bad IDs that exceed the 65,536 cap.

We have IDs using 100000-100025 that should be redone to valid numbers.

We may also want to consider a re-org of IDs to have them categorized as trying to just increment after last ID is painful when the IDs are shared between all .json files
![image](https://github.com/user-attachments/assets/db1300d2-4259-42d1-8a97-92e204390ef2)

Doxorn:
I think I could write a python script to to go through all the files and reassign new unique key values across all files to all entries according to set rules. Is the above list of files containing unique key values comprehensive?

CollinHerber:
I'm a bit confused by this. I've definitely seen the IDs that are in the high numbers work. Are we sure we aren't talking about actual entry limits and not the numbers themselves?

Bakelt:
Ran a quick test. Setting one text to the %65536 of the other causes the second string to no longer work. Currently, our 100k+ string IDs are only working because nothing is using their %65536 version.

Having grouped IDs would also be nice for making new tooltips in general. Ran into an issue with these before as the next string ID was being used an entirely different .json
![image](https://github.com/user-attachments/assets/aa5f6289-d53d-49eb-a16f-8bb026b3a58b)
![image](https://github.com/user-attachments/assets/bd3912bb-6890-44be-8315-393b94007b99)

Bakelt:
> I think I could write a python script to to go through all the files and reassign new unique key values across all files to all entries according to set rules. Is the above list of files containing unique key values comprehensive?

The main hurdle I see is that we likely don't touch half the .json files while Bonesy does. So we may need to take a look to see if those are sequential or not or if we'd end up having to just work around them.

I also don't know if we need to touch the other abnormal .json files like thul_rune_stacked.json

Doxorn:
> > I think I could write a python script to to go through all the files and reassign new unique key values across all files to all entries according to set rules. Is the above list of files containing unique key values comprehensive?
> 
> The main hurdle I see is that we likely don't touch half the .json files while Bonesy does. So we may need to take a look to see if those are sequential or not or if we'd end up having to just work around them.
> 
> I also don't know if we need to touch the other abnormal .json files like thul_rune_stacked.json

Would a script that collects unused ID ranges for each file work for you?

Bakelt:
[...]

PR #423, opened 2024-12-30
fix: ID corrections and script

CollinHerber:
Resolves: #265


Delegus (changes requested):
Breaks multiple vanilla ID's turns much of the UI is turned into missing string.

Bakelt (approved):
LGTM. Previously broken strings are all ok.
image
image

Bakelt (approved):

PR #438, opened 2025-01-02
Meteor impact delay

Delegus:
Meteor launch sounds are not in the casc and I am unable to edit so they are set to volume 0. Meteor targeting particles are not editable for timing and removing references results in lighting glitches so a new particle has been used in it's place.

Timing for initial delay reduced to 1.20 seconds from 2.4 seconds from cast to impact. Fire cast sound still plays, impact sound still plays, just removed the in between sound delay and extra unity particles.


Bakelt (approved):
LGTM

PR #441, opened 2025-01-04
Treasure Class Updates

### LINKED ISSUE #428: Gem Drop Chances
Reduce Gem drop chances further given credits are now universal.

Also need to update Andariel's loot tables in Normal and Nightmare to make sure she can't drop Chipped/Flawed/Flawless gems any more.

Furthermore, as part of the gem rework some items were removed from TC which makes drop chances much higher for things like jewelry - need to readjust these drop rates to bring them back in-line.

Vingelis:
Resolves #428

Gem Drop Chances, Andariel Loot Table Fix, and Jewelry Chance tweaking.

Conversion Orb picks reduced back to 1.


Delegus (approved):
Tested and feels good to go.

PR #450, opened 2025-01-08
Adds Cube Loot Skill for Controller

Delegus:
Adds the Cube Loot skill and makes it functional for controller users.

### Preview Builds Only
- Act 3 mercenaries will need to be rehired.
- Metamorphosis runeword will need to be re-crafted.
- Splash Charm will need to be re-found and upgraded again.

PR #455, opened 2025-01-09
Frozen Orb

Delegus:
Very slight speed increase to the orb. Top end damage increased in values of 2 to make higher skill values feel more impactful.


Bakelt:
Will conflict with #450. Recommended we resolve the conflict there and then merge #455 after due to how small this change is

Bakelt (approved):
LGTM

PR #456, opened 2025-01-09
Arrow and Bolt

Delegus:
Removes random arrow and bolt craft recipes that can only spawn as white items with additional modifiers.
Updates rep-quant unique items to be universally 1 per 4 seconds.

Added in 2 unique arrow and 2 unique bolt recipes that require level 85 to equip and cast a curse on hit.
image
image
image
image

Crafting Recipes
image
image


Bakelt:
I think the %chance should be lower than 100% for visual and sound clarity if nothing else. Maybe 25%?

Bakelt (approved):
LGTM

PR #458, opened 2025-01-09
BugFix-RandomSummonAuras

Bakelt:
- Moved the summon splash from monequip to monprop as the Summon Armor was rolling random auras
- Removed Inventory flag from necro and druid summons since they use monprop now


Delegus (approved):
LGTM

PR #459, opened 2025-01-09
Uniques-SpellingCorrections

Bakelt:
- Fixed a few incorrect ItemNames

Note: This is nonfunctional but will make some data parsing of the file easier


Delegus (approved):
lgtm

PR #462, opened 2025-01-11
Update charm weight text

Delegus:
Retroactively updates, no save break.
Changes text on the +1 Activates Charm Weight to +1 Charm Weight Active; This is the best I can do with my knowledge of the game files as I cannot find a way to make it an entirely insivisible property.
image

oskills on other items.
image
image
 This change is already in 2.0


Vingelis:
Let’s just get rid of ‘active’ and just have it say ‘+1 Charm Weight’.

Bakelt:
oskill is apparently a hardcoded property so it can't be cloned and used without a description (unless you nuke all oskill descriptions)

I prefer the "Charm Weight Active" over just "Charm Weight" to make it more distinguished from a charm's weight

Bakelt (approved):
LGTM

PR #463, opened 2025-01-11
MinManaSettings

Bakelt:
Resolves: #442
- Min Mana for skills set to at least 1 so that you can't bypass charm weight with blood mana

Bone spear uncastable despite blood mana being active
image


Delegus:
Tooltip for magic arrow still shows no mana cost with enough points but it does still require 1 mana to cast. Dunno how many skills cost less per point like it does, do we want the tooltips updated now to reflect properly or add as an issue and fix later?

Bakelt:
I think fix later as spells costing 1 mana instead of 0 should be pretty negligible compared to being able to bypass the charm weight

Delegus (approved):
LGTM on implementation, issue #464 opened for tooltips.

PR #472, opened 2025-01-12
Assassin Tweaks

Bakelt:
### Blade Sentinel
- Blade Sentinel can now fly so it won't have to path around terrain
- LocalDelay removed. The skill still has a cap of 5 sentinels
- Mana cost changed from flat 7 mana to a scaling curve.
- Move speed increased by 50%
- Can now hit every 15 frames (was 25)
- Added flat AR and AR/lvl similar to other physical skills

### Blade Shield
- Mana cost reduced
- Delay between hits reduced
- Can now hit every 15 frames (was 25)
- Added flat AR and AR/lvl similar to other physical skills

### Shadow Warrior
- Mana cost reduced 

### Shadow Master
- Mana cost reduced


Vingelis (approved):
LGTM

PR #481, opened 2025-01-13
Orb of Shadows

### LINKED ISSUE #479: Orb of Shadows
Add a new extremely rare orb that allows players to add Ethereal status to an item

Vingelis:
Fixes #479


Vingelis:
Need to either update cube recipe to make items Eth (change material requirements) or remove it.

Vingelis:
Will eventually introduce a new type of orb that extracts an essence from an existing Eth item and then used as part of an updated cube recipe to make an item eth as another route for players.

Bakelt (approved):
LGTM - we can revise drop rates later if they're not as intended

PR #499, opened 2025-01-16
Souls skills

Delegus:
Set souls on their own skill so changes to the sorceress chain lightning do not apply to their damage.

Testing shows they still hit hard just not with every single shot like before.

Resolves: #469


Delegus:
Further testing and investigation reveals I was wrong how this works and this is an invalid change.

PR #503, opened 2025-01-17
HB FotH Updates

Delegus:
So far updating HB and FoTH so they are more in line with each other. Still trying to figure out the -MDR for paladin and can't get it to work on any aura that isn't offensive due to coding limitations they apply to you instead of the enemies.

Holy Bolt
- Synergies reduced to 25% from 50%

Fist of the Heavens
- Lightning Damage synergy up to 15% from 7% from Holy Shock
- Holy Bolt Damage synergy up to 20% from 15% from Holy Bolt
- Cooldown Normalized to match tooltip and other CD skills in 2.0 (Now 0.6s CD down from 1s)


Bakelt:
Yes, you can’t buff yourself and debuff enemies. I’d recommend Sanctuary as it’s anti used.

Delegus:
> Yes, you can’t buff yourself and debuff enemies. I’d recommend Sanctuary as it’s anti used.

This is counterproductive to the goal of making hammers feel better as HB ignores all monster resistances and FotH doesn't gain a bonus from concentration. Will need to either buff hammers at higher levels or buff conc damage increase.

Bakelt:
Have you tested HB against magic resistant / immune enemies? Can easily modify Fallen to be immune and verify if it’s doing Magic damage

Delegus:
Yeah looks like ignore flag was removed or info I've found on vanilla is wrong. Testing shows the bolts are both affects by magic resistance. Sanctuary at the moment supposedly only affects undead enemies, so in order to be useful it would need to affect all enemies and then conviction and concentration would no longer be used. Kinda stumped on the -MDR thing for now. May just need to tell people to get a CtC weaken item or add to some paladin items or a random affix on paladin shields.

Bakelt:
-MDR is the easiest one to not have many sources for as the pool of enemies at 99 res is very small. If we were to give an aura source to pala, I find it better design to expect people to swap auras when they are at high MDR enemies.

We can always make "-MDR" items by making a new aura that reduces MDR for nearby enemies.

Delegus:
Will revisit paladin in the future for now HB and FotH stuff.

Vingelis (approved):
LGTM

PR #504, opened 2025-01-17
Add Gem Bag to Starting Characters

### LINKED ISSUE #500: Add Gem Bag to starting character
Add Gem Bag to new character inventory.

Vingelis:
Closes #500 

**SAVE BREAKING CHANGE: throw away Gem Bags and Keychains before updating.**

Reworked Gem Bag to be distinct from keychain storage.

Reworked Keychain IDs to be consistent. Now also properly references its assets files.

Added Gem Bag to all new starting characters.

Added Gem Bag and Keychain to vendors for an appropriate price.

Removed Cube recipe to create Gem Bags and Keychains.

Reworked tooltips on both Gem Bag and Keychain to be more intuitive.

Added flavor text to Gem Bag, Keychain and Infernal Keys.

Reduced vendor buy prices for gems (30,000 -> 1000)


Delegus (approved):
I've gone through this. It isn't save breaking just item breaking. Old bags and chains no longer function. Save breaking is when your chars won't load. That being said testing looks good.

PR #506, opened 2025-01-18
WeaponDependentSkills

Bakelt:
- Tectonic Slam charges will no longer spawn on non-Slam weapons
- Poison Dagger charges will no longer spawn on non-Knife weapons
- "Epee of Speed" unique now has charges of Concentrate instead of Stun/Tectonic Slam as it is not a slam weapon

Note: Double check recent keychain/gembag change was not blown out


Delegus (approved):
Doesn't look like anything is amiss with the GitHub changes, no changes to anything already modified and it is merged into next.

PR #512, opened 2025-01-19
add rune table to ingame menu

The-Honorable:
updating tab2 of modinfo to contain a table of the runes/what they do to avoid tabbing out to hit the wiki. what it looks like below.

image


Vingelis (approved):

Vingelis (changes requested):
- Remove 'To' anywhere it's used
- Damage -> Dmg
- Against -> Vs.
- +2 Mana After Each Kill -> +2 Mana Per Kill
- +75 Poison Dmg Over 5 Seconds -> +75 Poison Dmg / 5s

The-Honorable:
updated. new example image
image

Vingelis (approved):
LGTM

PR #526, opened 2025-01-23
Sorcy Armor and Warmth

Delegus:
FotH Holy Bolt NHD Activated, not sure why it was removed. One FotH on a mob causes all of the HB to murder the target currently. Not sorceress related but not worth making a separate PR for.

Goal is to give incentive hard points and add a little flexibility in going for breakpoints without removing the vanilla utility.
- All 3 armor 1% base damage reduction +1% damage reduction per hard point with a max of 25
- Warmth +5 life after each kill base + 1 per hard point with a max of 30
- Frozen Orb got another slight boost to top end damage spread because it still just felt bad in hell.

Warmth:
image

Armor:
image

Projectiles Tooltips Updated:
image


Delegus:
Updated Warmth to be 0-30 instead of -1 to 29.

Delegus:
Tooltips are the sticking point for me. Want to either make it all synergy description like the + projectiles or make the + projectile built in like the + PDR and Life after each kill in the armors and warmth.

Delegus:
Done tweaking

Bakelt (approved):
LGTM

PR #537, opened 2025-01-25
WP UI Animated And More TZ Updates

Delegus:
Waypoint button sprites updated so that when you click them there is an animation and the icons and text no longer disappear.
- Nothing selected
image
- Spider Forest Selected (Can't SS cursor in sadly.)
image

WPs updated to include Terror Zone Icon and Text coloring to indicate Terror Zone
- Redundant I know, but Alice and her white rabbit beckoned, it's a simple flag change to disable if unpopular.
- Created a Terror Zone Sprite out of the Vanilla sprite to match our waypoint icon.
image

desecratedzones.json re-ordered so that zones are displayed in order by adjacency to other zones instead of by level ID.
- Multiple waypoint ID's added/fixed in order to make the shiny new sprite and text work.
- Den of evil replaced by Dark Woods because Den was lonely and didn't make sense.
- Black Marsh re-added cause it looks like it was removed by accident.
- Canyon of the Magi added because there is space in A2 for it.
- Sewers Level 2 added because there is space in A3 for it.

### DIFF
--- a/data/global/ui/layouts/controller/waypointspanelexpansionhd.json
+++ b/data/global/ui/layouts/controller/waypointspanelexpansionhd.json
@@ -3,16 +3,16 @@
     "type": "WaypointsPanel", "name": "WaypointsPanelExpansion",
     "children": [
         {
-            "type": "ClickCatcherWidget", "name": "ClickCatcher",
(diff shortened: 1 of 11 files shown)


Bakelt (approved):
LGTM

PR #542, opened 2025-01-25
switching for oskill to skill for druid specific set

The-Honorable:
fixes issue https://github.com/D2R-Reimagined/d2r-reimagined-mod/issues/535

switched oskill to skill, and capitalized skill names because it matter/crashes the game otherwise.

image

22 werewolf before set bonus
image

30 after
image


Vingelis (approved):
LGTM

Issue #517, opened 2025-01-20
Paladin Late-game HolyFire Nerfs


Bakelt:
Pala HolyFire builds were a problem in v1.21 and needs some toning down.

1. Lower synergy values as this won't impact it as a leveling build
2. Cap the aura range so it's not hitting things 2 screens away (should do the same with holy lightning/freeze)

PR #545, opened 2025-01-26
limit holy auras to 30 yard, nerf holy fire synergies

The-Honorable:
Resolves: #517

### Holy Fire
- aura range capped to 30
- synergies reduced to 21 resist fire/10 salvation -> 16/8
image

### Holy Freeze
- aura range capped to 30

### Holy Shock
- aura range capped to 30


Delegus (approved):
Looks solid.

Issue #544, opened 2025-01-25
Vengeance Paladin Set Idea (Supporter: LickMeImMelting)


Vingelis:
Idea for Vengeance Paladin Set from supporter LickMeImMelting

Weapon: war hammer  Req lvl 18

+2 min/max damage per level ( starting sort of like a civerbs, but ending like a grief.)
100% ED 
20% IAS
+x to Vengeance

Set bonus:

2- set bonus:IAS 40%?

Complete set bonus: + 50 min/max damage?

Paladin shield. Req Lvl 50
+45 all res.
FHR
Faster block rate

Set bonus
2- set bonus: Damage reduced by 15%
Complete set bonus: Level 12 Holy freeze/Salvation aura when equipped

belt req Lvl: 30?

+20 to fire/light/cold res.
Damgage taken goes to mana.? 
30 mf?

Set bonus: 
2- set bonus: +30 fire/light/cold res

Complete set bonus +40 fire/light/cold res.

lickMeimMelting:
Ive been told that if we do the set bonus +50 min/max damage it will automatically change to 50% ED so to change this it might be smarter to go for +40 min/ +50 max damage as the set bonus. Not sure if this would take up 1 or 2 slots though

lickMeimMelting:
**Weapon:** 

+2 min/max damage per character level.
!00% Enhanced Damage.
+20% Increased Attack Speed
+3 to Vengeance

**2-piece set bonus**
+ 40% increased attack speed.

**Belt**

+ 20 Fire Resistance
+20 Lightning Resistance.
+20 Cold Resistance
+30% chance to find magical items.

**2-piece set bonus**
+ 30 Fire Resistance
+30 Lightning Resistance.

**Shield**

+45 to all resistances ( standard paladin shield bonus)
+45% Faster Hit Recovery
+55% Faster block rate
Level 12 Holy Freeze Aura when equipped.


**2- set piece bonus**

+15% damage reduction
+30 Cold Resistance

**Complete set bonus**
+40 minimum damage
+50 maximum damage
+40 to all Resistances

PR #547, opened 2025-01-27
Melting Vengeance Set

Delegus:
Resolves: #544

itemstatcost.txt
- Added stat minimum damage per level and it is calculated in 8ths same as other per level damage stats.

properties.txt
- Added a property to combine min/max damage per level into a single property for my sanity.

3 piece set centered around the skill Vengeance that is intended to be useful throughout the game.
Weapon and belt can be used together on other classes so it may also find a niche as a leveling set for melee classes from 40+

image
Weapon requires level 20, drops at level 12 and onwards. (Andy Normal +)

image
Belt requires and drops at level 40 (Baal Normal +)

image
Shield requires 60, drops at level 59+ (Mephisto NM +)


Bakelt:
Haven't run D2R lint but everything else checked out fine.

Vingelis (approved):
If he happy, we happy

PR #548, opened 2025-01-27
Remove charm weight text

Delegus:
NOTE: This will cause all items with oskills found in the 2.0 preview patches to no longer show those oskills as a string, but the skills are still there and still apply to the character. Item will need to be found again to display the oskill in the text again.

Removes +1 to Charm Weight Active Text from all weapons and shields though the property is still there but now hidden.
- Text is not removed from the .json file in case we do want it visible on something in the future and to prevent an ID merge issue since it is 5 or so entries up.

Reset oskill string display position to the original value so they show up where they used to on items.
Added Charm Weight to new Maadi Set weapon and shield.

Charm Weight Applying from Weapon - No Text
image

No Weapon Same Char
image


data/global/excel/itemstatcost.txt
```diff
@@ -386,5 +386,4 @@ item_splashonhit_five	383		1	7	16						1				-400	0		21	0	7	0	16										1	domel
 charm_weight	384		1	7																7	6																1	19		mod_weight	mod_weight									1	0
 mana_control	385																			6				11			mana										1	19		weightsystem	weightsystem									1	0
 stacked_gem	386		0	16								1	0	0	1	0		16	0	16	0																56	19	0	StackedGem	StackedGem										0
-item_mindamage_perlvl	387		1	6											54	204		6	0	6	0			4	3	level	mindamage	secondary_mindamage	item_throw_mindamage			1					126	19		ModStr1g	ModStr1g	increaseswithplaylevelX									0
```
Delegus: Removed new stat added in Melting Set PR

data/global/excel/setitems.txt
```diff
@@ -424,5 +424,5 @@ Temptation's Death	420	Corruption Coils	amu	Amulet	1	45	45								5	5000	2	hp		5
...
 Maadi's Vision	422	Maadi's Paradox	uh9	Bone Visage	7	63	63								5	5000	2	ac%		150	200	vit		25	25	balance2		20	30	res-all		10	10	mag%		30	50																	move3		15	20					cast2		15	15																														0
 Maadi's Spirit	423	Maadi's Paradox	uhn	Boneweave	7	47	47								5	5000	2	ac%		150	200	vit		50	50	enr		50	50	hp%		10	15	ac		100	200																	vit		50	50					enr		50	50					abs-mag%		10	10					red-mag		10	10														0
-Maadi's Torch	424	Maadi's Paradox	7gw	unearthed wand	7	64	64								5	5000	2	allskills		1	2	extra_bonespears		1	1	cast2		30	30	mag%		10	30																					cast2		25	25					res-all		20	20																														0
-Maadi's Soul	425	Maadi's Paradox	ned	Overseer Skull	7	49	49								5	5000	2	allskills		1	2	cast2		20	30	res-all		20	20	block		20	50	block2		20	50	ac%		150	200																					skill	BloodGolem	5	5					skill	Golem Mastery	5	5					extra_bonespears		1	1														0
+Maadi's Torch	424	Maadi's Paradox	7gw	Unearthed Wand	7	64	64								5	5000	2	allskills		1	2	extra_bonespears		1	1	cast2		30	30	mag%		10	30	oskill_hide	393	1	1																	cast2		25	25					res-all		20	20																														0
```
Delegus, line 427: Adds Charm Weight to Maadi Set

Delegus:
Tired of merge conflicts with this, let's test and push it in please.

Vingelis (approved):
LGTM

PR #550, opened 2025-01-27
removing stamdrain on all items

The-Honorable:
Resolves: #549 

# Remove stamdrain affix from uniques/rune/runewords/set items where it was present.

# add life after each kill to ELD in place of stamina drain

swapped stamdrain to heal-kill on eld. is 5 lpk too much?

image

image


Vingelis (approved):
LGTM

PR #555, opened 2025-01-28
Fix Succubus Witch Skills

### LINKED ISSUE #554: Fix Succubus Witch Skills
Succubus-type minions had Blood Mana skill removed and SuccubusBolt moved up in its place. We need to move the SuccubusBolt skill back to its original place or remove the MonAI entry for Blood Mana

Vingelis:
Fixes #554


Delegus (approved):
Solid

PR #558, opened 2025-01-29
add in initial raven set for testing

The-Honorable:
# New Druid Set + tweaks to ravens

## Raven Tweaks
- Raven velocity changes up by 5
- gains slight attack speed with each level

The Raven's Nest 5 Piece set
  - End game Summon Druid Set geared towards using ravens as a primary damage source while limiting wolf summons capacity

## example images of set
image

image

image

image

image


Delegus:
Initial is solid, but removes all of the "Maadi's Set" from item-names.json, also no changes to skills.txt so the adjustments to max summons and raven damage aren't applying.

The-Honorable:
> Initial is solid, but removes all of the "Maadi's Set" from item-names.json, also no changes to skills.txt so the adjustments to max summons and raven damage aren't applying.

yea fixed the bad commits/merged. holding off on the skill changes as a commit until i understand the changes you and Bakeit were proposing and we can get it right. the itemstatcost and property stuff im not yet familiar with.

Delegus (approved):
Caw

PR #574, opened 2025-02-02
setting skills to have min mana in description/usage

### LINKED ISSUE #464: Investigate Mana Tooltips for the Minmana change
See change implemented in #463 
Some mana costs are now min1 but tooltip will show 0 (Magic Arrow, Double Swing, Sacrifice)

The-Honorable:
Resolves #464 

# Magic arrow

image

# Double Swing
image

# Sacrifice
image

also without any mana you cannot use it, but it does not consume any mana
image


Delegus (commented):
Since Sacrifice doesn't cost mana, I would definitely remove that from the string. Just having it not work if your mana is set to 0 by charm weight is good enough. Tested and looks good to go otherwise.

The-Honorable:
updated so the mana cost is not present on tooltip

image

Delegus (approved):
LGTM

PR #576, opened 2025-02-02
update 1 charge uniques

The-Honorable:
Resolves: #444 

# Voice of the prophet 1 charge of Battle command/orders/shout -> 10
image

# Killer's Watch 1 charge of level 6 iron golem -> 5
image

# Doom Avatar 1 charge of level 50 revive -> 5
image


The-Honorable:
flipped to ready. since the 2 uniques are summoning i think it makes sense to keep it to like 5max. shouts you are likely to use more than once in a session though.

if there are any reasons these should not be changed or are intentionally left to 1 let me know.

Delegus (approved):
LGTM

PR #581, opened 2025-02-03
Potions setfunc fix/ Xp cap

KingPandaYumYum:
setfunc set to 1

xp pot tooltip needs to be changed as 255% is the xp cap


Delegus (approved):
LGTM

PR #584, opened 2025-02-03
Vendor Panel Confirm Removal

Delegus:
This update removes the Yes/No confirmation from vendor panels, it just autoselects yes with no delay so you can fill your inventory with a left click macro with potions/gambling, etc.

Side Effects: Removes "Not Enough Gold" from buying items or repairing as a clickable confirmation but it is pretty obvious when a vendor action fails due to a lack of gold. Item doesn't appear in inventory, repair doesn't happen and Repair All cost stays the same.

https://github.com/user-attachments/assets/06b67faa-568c-49e1-a117-617e74718908


Vingelis (approved):
LGTM

PR #585, opened 2025-02-03
Upgrade Item System (A.K.A Add Stats) & other Misc changes

### LINKED ISSUE #179: Stat overflow for Item upgrading via Cube

### LINKED ISSUE #356: Rebalance "Add Skills to Amulet" Recipe
Spawned from #350. Recipe is too cheap for the benefit.

### LINKED ISSUE #361: Hellfire Torch Rework
**Problem:** 
Hellfire Torches are seen as a staple for all end-game characters, but are painful to obtain for several reasons:
- Key farming is down to RNG. Requiring 3x3 keys can take a good while to find without a recipe to convert keys
- Frustrations due to disconnects/deaths meaning opening portals again and potentially having duplicate reagents 
- Ubers require specific classes/builds to beat consistently, which is counter to idea of allowing all classes/skill trees to flourish in a single player setting
- 1/7 chance of rolling your class
- Stats have a large variable window
- On-hit effect is visually distracting and does nothing

**Proposal:** 
Remove the existing Hellfire Torch.

Introduce a new 'torch' that works similar to Splash Charms. Players would obtain the first level (Small Charm) when killing normal Baal. The charm is upgradeable to Large and Grand Charm levels later on.

- The Small Charm would provide generic stats and +All Skills that would be beneficial for levelling.
- The Large Charm adds to this, as well as providing class/build specific stats intended to help get through later difficulties. 
- The Grand Charm culminates in BIS stats for classes/builds, and is intended to help players with end-game content.
- The 'Uber' Grand Charm is the final form.

Reagents required to upgrade would be suitably expensive for the benefit they provide, and should be naturally obtainable after some farming. Killing Nightmare Baal and Hell Baal would provide **one** of the required reagents to upgrade to Large and Grand Charms respectively. Ubers would provide the final reagent needed to upgrade to the 'Uber' Grand Charm.

The Large and Grand Charms will provide benefits for different builds (i.e. Light/Fire/Cold Sorc, Shape-Shifting, Elemental or Summoner Druid, etc.). When upgrading from Small -> Large, or Large -> Grand, players would choose which of the three skill variants they want by using an appropriate recipe. Players could swap between build-specific variants using the cube and an appropriate set of reagents, or just cubing the charm alone (whatever makes the most sense balance-wise).

**Outcome(s):**
Provides gradual progression for players that will help with the revamped difficulty being proposed, and reworks the existing torch into something that properly benefits players instead of just being a stat stick. Player agency is improved by allowing them to swap between class builds and still get benefits from their torch. Ubers dropping a reagent instead of a torch keeps them relevant, alongside the standard of heroes, and potentially anything else we can think of.

### LINKED ISSUE #432: Add Item Stats Changes
Look to restrict Add Item Stats recipes to X many per item to prevent ballooning both of save file sizes and of too much power.

### LINKED ISSUE #470: Charsi Imbue / Akara Respec Loops
1) Charsi Imbue quest becomes somewhat obsolete with the intro of Orb of Infusion. Is there a way to make her reward repeatable in some way? If not, consider making Orb of Infusion into a purchasable object from Charsi/Farah/Hratli/Halbu/Larzuk instead of being found in the wild. Or, still have both, but make Orb of Infusion more rare, and make the cost of the purchasable object competitively priced (can this be based on character level?)

2) Akara Respec is obsolete with current TP + ID Scroll recipe. Consider Akara/Lysander/Alkor/Jamella/Malah selling Token of Absolution instead, for a competitive price (based on character level?).

Another idea, consider introducing rare reagents as part of purchase price for various items we want to start offering in stores.

### LINKED ISSUE #553: New recipes
I think that in this version, in endgame,  magic jewels have become more important than rare or unique ones (lower level requirements).
So I think we should add :
A) Recipes to get magic jewels from rare or unique ones.
or B) Recipes to lower level requirements using those rare or unique jewels with more common runes than Zod.
For example a more common rune with 2 rare jewels and and even more common one with one unique jewel.

### LINKED ISSUE #598: Move Throw Weapons from Malah to Larzuk
Move Throw Weapons from Malah to Larzuk

### LINKED ISSUE #601: Add Unsocket Items
Add new items for unsocketing runewords and non-runeword items - make them purchasable from vendors

Vingelis:
Fixes #179 
Fixes #356 
Fixes #432
Fixes #553 

Introduction of upgrade caps per item type (2 for Jewelry, 3 for helms/belts/boots/gloves, 5 for 1h weapons, shields, chest, 10 for Slam weapons, Spears/Polearms and Bows/Crossbows).

All recipes have been reworked given the limited number of upgrades per item slot.

Removing recipes for stats/lvl and reduced vendor prices because they break things too much.

Fixes #361 
Introduces Hellfire Torch -> Obsidian Beacon, and Annihilus -> Black Soulstone upgrade paths. Trying to carry more than one Torch, or more than one Soulstone at a time has a similar effect to carrying 40+ Charm Weight, and sets player HP to 1.

Hellfire Torch no longer drops from Uber bosses. It is now a craftable charm that requires three unique reagents that are dropped by each of the Uber bosses.

Fixes #470 
Added Token of Absolution to potion vendors in all acts - changed to itemtype 'elix'. Removed TP + ID Scroll cube recipe.

Fixes #598
Moved throw weapons from Malah's Misc. tab to Larzuk's Misc. tab

Fixes #601 
Added Rune Pliers & Jewel Pliers for unsocketing runeword and non-runeword items. Made purchasable from Blacksmith vendors in all acts. Removed TP + Item cube recipe for unsocketing.

Misc. Changes:
Low/Mid tier Item Crafting Recipes changed to require a single gem of any kind instead of using the gem bag. Output item now always uses usetype, and sets lvl to 100% plvl instead of a mix of plvl and ilvl. Also removed character level requirement for all crafts. Will revisit item crafting recipes altogether post 2.0.

Removed Stamina potions from all vendors

Updated rune/rune stack names to change (X) to (#X) to help clarify that it represents the rune's number, not quantity.

Updated many item property strings to be more descriptive.


Vingelis:
2h weapons (miss, slam, sppl)
Single resistances
Akara Respec token
Unsocket items
Move thrown weapons from Malah to Larzuk

Delegus (approved):
Changes reviewed over the past week in stages and we've fixed, or attempted to, all the major and minor issues.

PR #588, opened 2025-02-04
fixing firecam chest/pull from darkness 

The-Honorable:
Resolves: #569 
Resolves: #561 

moved half the res from flat roll to 4 set bonus. flat on chest now rolls from 75-100 -> 30-50
also added 20 fcr to the chest so its slightly more useful as a standalone/in the set.

image

switched pull from darkness bone spirit roll from 7-15 -> 5-7 to be within bounds of actual rolls since 7 is max.

image


The-Honorable:
ready for review :)

PR #590, opened 2025-02-05
Necro mage synergy rework

Bakelt:
- Necro Mage's missile skill level is now purely based on the level of raise skeletal mage
- Necro Mage's now inherit sunders for phys, cold, fire, and lightning
- Max Mage's set to 14

image

Balancing:
- Synergy parameters can be adjusted at will and will be auto updated on the end value
- Skill curve changes require changing passivecalc12-14 (ps12 = fire/cold/phys min, ps13 = fire/cold/phys max, ps14 = lightning max)
- Currently have enough free passive calcs (6) to split up fire/cold/phys if needed


Delegus (approved):
Merging so I can tweak synergies on my own PR as I also intend to tweak Telekinesis testing shows the tooltips work properly and display the proper values. Synergies work properly and provide the right damage increases.

PR #591, opened 2025-02-05
add mf section to breakpoints

The-Honorable:
image


Delegus (changes requested):
Since it says 400 - 600 either make that 500% or add the row for 600% please.

The-Honorable:
updated line 3 words and added mf600

image

Delegus (approved):

PR #592, opened 2025-02-05
Necro Mage and Warrior Synergies - Potion Tweak

Delegus:
Resolves: #510 
Related: #590 

Character select mod version updated.

Raise Skeleton
- Synergy with Skeleton mage now 10 from 3, this synergy only starts to apply around hard skill levels 15+ so helps with the NM skill damage fall off and brings them up a bit at higher skill levels to be competitive in hell.

Raise Skeleton Mage
- Skeleton Mastery Synergy is now 5 from 8 due to being soft points, the value gained is consistent to max skill level.
- Skeleton Warrior Synergy is now 10 from 12 to even out the early mid damage boost from hard pointing warriors.

Health and Mana Potions
- T1 %1/4
- T2 % 1/4
- T3 % 1/3
- T4 % 1/3
- T5 % 1/2

Assassin
- Made all uses of charges cost only 1 charge

Panda Potions
- Description updated to warn that duration stacking doesn't work.


Vingelis:
Potion changes feel good. We may want to pull back very slightly on the MP and HP TC changes from prior PR.

Assassin changes look good too. This will mean runewords like Mosaic is less of a necessity.

Will leave it to others to comment on Necromancer changes, no idea about balancing those skills.

Bakelt (approved):
LGTM

PR #596, opened 2025-02-07
add in old bens walking stick runeword

The-Honorable:
# old ben's walking stick

meant to be a leveling runeword to speed through the first bit of the normal. comes with downsides that would dissuade you from using this in nightmare/hell. -drain life and -gold/mf.

# level 7 stats 
### normal hp1 do barely out heal it at level 7

image

# level 100 stats
### drain is roughly 73hp/s which is basically the max it can be set

image


The-Honorable:
anyone wanna test this? :)

Delegus (approved):
Works as expected

PR #611, opened 2025-02-11
Solar Creeper

Delegus:
Gave Creeper a floor AoE damage on hitting an enemy so it does something until a better rework is performed.


The-Honorable (approved):
lgtm

PR #626, opened 2025-02-15
ItemUpgrade-Helm-AddEleSkil

Bakelt:
- Fixed the helm upgrade recipes all defaulting to +Fire Skills

Would like to test that it still works with existing +Fire Skills items as the string is removed from item_elemskill but don't currently have any handy
image


Delegus (approved):
Prior upgrades tooltip won't show any longer, thankfully still caught it early. Works as intended with a side effect that skills.txt etypes need updating in a couple of spots.

PR #634, opened 2025-02-15
LanguageFix

Bakelt:
- Missing languages were breaking the .jsons for several languages


Delegus (approved):
Could've sworn I had him fix this on the original pull request, and then I fixed it again on the one after when Ben and I were getting Raven set in. Oh well.

Issue #640, opened 2025-02-16
AffixNerf-RingFlatReduction


Bakelt:
![Image](https://github.com/user-attachments/assets/01d50900-8ea5-44b2-b797-1098e1a7e999)

Issue #639, opened 2025-02-16
ItemNerf-EcstasyOfIshtar


Bakelt:
Reduce Max HP/Life
Reduce flat -phys/-mag
Piercing is probably fine
![Image](https://github.com/user-attachments/assets/c0faedba-0873-4477-b502-684779b10771)

PR #641, opened 2025-02-16
ItemNerfs

Bakelt:
Resolves: #639
Resolves: #640

- Reduced the %MaxLife and %MaxLife on Ecstacy of Ishtar from 25% to 10%
- Reduced the flat reduction on Ecstacy of Ishtar from 25-30 to 10-15
- Ring's can no longer roll the "of Life Everlasting" suffix (highest flat reduction tier).


The-Honorable:
0.0 25-> 10 is a lot. not 15? could also raise the rarity of the belt?

Bakelt:
The big draw of the item is the 50% pierce. Even with the nerfs, it is probably one of the best defensive belts. I was using it on my melee barb because the stats are gross.

PR #657, opened 2025-02-18
2.0.2 Mini TC Update

### LINKED ISSUE #629: TZ Rune Bosses
Our changes to the Rune bosses no longer extend to their terrorized variants though we discussed this extensively and it was fixed at one point during testing, need to check for a merge that messed up the TC changes.

Delegus:
Resolves #629 

treasureclassesex.txt
- Add slight NoDrop to socket and corruption 
- Sunder charms reverted to vanilla NoDrop
- Fix Bloodwitch NM and H TZ picks to drop runes properly
- Fix Battlemaiden NM TZ picks to drop runes properly
- Fix summoner key drops to be consistent with the other 2 key bosses and across TZ
- Fixed Hephasto runes to just be 2 always and fixed his NM picks in TZ
- Fix Nihlathak TZ key probs to be consistent with non-TZ
- Fix Frozenstein runes to just be 2 always, and his NM picks in TZ
- Fix Countess TZ key probs to be consistent with non-TZ

superuniques.txt
- Updates LT mobs to have loot drops from unique tables due to community feedback about lack of drops.


Bakelt (approved):
LGTM - Loot seems fine. Not seeing shadow orbs drop like candy. Will adjust Raven Talon in another PR

PR #667, opened 2025-02-20
TC Minor Fix

Delegus:
Minor TC fix while I work on the major one.

Dial back ND of socket and corruption orbs.
Stick sunders to a more reasonable drop rate attempt 3.


Delegus:
Consensus to finish overhaul and not bother with this.

PR #676, opened 2025-02-21
Increase Battlecry Range

Bakelt:
- Increased Battle Cry Range Note: Visual on HD graphics isn't accurate. Will hopefully be able to resolve later


The-Honorable (approved):
LGTM

PR #679, opened 2025-02-21
add flying scim json to adjust scale of swords

The-Honorable:
set to 1.8, could maybe be lowered to 1.5 ish. 2.0 seems maybe a tad too big.
unable to scale the particle from my limited testing.


The-Honorable:
also re-enabled constricting ring
image

Bakelt (approved):
We may want to revisit the swords but I think this is sufficient for an initial change.

PR #680, opened 2025-02-21
Feature/fix tiger eye

The-Honorable:
- fixed tiger eye unique jewel name color from blue -> gold to match other unique jewels

image


Bakelt (approved):
LGTM

PR #683, opened 2025-02-22
New unique quiver repairs

Delegus:
- Adds recipes to repair new unique quivers
  - Testing shows this value wraps around starting at 256 so 255 is the max value that can be used with qty
  - It's only a HP potion so not worried about it

Updated quiver and bolt types in misc.txt to be 500 quantity and spawn with 500 on creation.

Updated absorb tooltips to have a + at the front.

Updated PDR tooltip to have a +/- at the front.


Bakelt (approved):
LGTM

PR #684, opened 2025-02-22
Fix Sanctuary On Equip

Delegus:
Removes mana/startmana/minmana fields from sanctuary so an on equip aura can properly activate on mercenaries.


Bakelt (approved):
LGTM

PR #696, opened 2025-02-23
Splashy and Vengy

Delegus:
Suffix "of the Lich" once again fixed.

Splash Charm reworked into a radial type splash that expands from the target struck instead of a cone from the character.
- No longer double dips source damage. 

Vengeance reworked to use the elemental skill damage fields instead of basing the damage off of the held weapon. Still carries leech, DB, CS, etc.
- Damage curve is shallow at first and ramps up past skill level 29 very quickly to facilitate hell gameplay.
- The skill damage now splashes properly with the splash charm
- Gave a minor physical bonus to facilitate leeching.
- Mana cost has been reduced.

Fire Claws has been adjusted to deal a firey explosion on hit.
- Skill damage does not splash with the splash charm, the splash is built into the skill.
- Damage curve adjusted to fit the new AoE role, but maximum curve is increased past skill level 29 to facilitate end game hell grinding.


Bakelt:
FYI, Sanctuary currently broken on your branch.
image

Bakelt:
Does this actually generate a missile w/o a missile call in the srvstfunc or srvdofunc?
image

Delegus:
Yep, I posted the video in contributors didn't I? Lemme repost.

PR #706, opened 2025-02-24
'Slower Stamina Drain' Rollable Affix

### LINKED ISSUE #664: 'Slower Stamina Drain' Rollable Affix
Rare and Magic Boots can roll the 'X% Slower Stamina Drain' as an affix, which is a dead affix here.
It would be nice to remove that from the pool, if possible.

Vingelis:
Fixes #664

Magic Suffix 'Of Traveling' gives stam drain reduction in addition to FRW. Without this affix its just a copy of the suffix 'Of Speed' with a higher level requirement. Disabling this as a spawnable affix.


Bakelt (approved):
LGTM

PR #712, opened 2025-02-26
add charge-noconsume to h2h

The-Honorable:


Delegus (changes requested):
Need to add a string for "of Mosaic" so when it spawns on a magic item it isn't showing missing string.

Vingelis:
Are you planning on changing item_charge_noconsume in itemstatcost so that it appears at the top of the affixes list, similar to +Projectiles?

The-Honorable:
> Are you planning on changing item_charge_noconsume in itemstatcost so that it appears at the top of the affixes list, similar to +Projectiles?

i wasnt, but i can.

Vingelis:
Think if you're going to change the text color, make it appear higher as a defining feature of that item.

Dys-the-Fox:
Would it also be a good idea to add this affix to all martial-based unique Claws so they aren't further invalidated by possible rares/rws?

Delegus:
Same issue as spear and holy bolt projectiles. Just moves 1 required item to another. Taking away from Mosaic is the goal with this, not making more of them.

The-Honorable:
> Need to add a string for "of Mosaic" so when it spawns on a magic item it isn't showing missing string.

should be fixed. dont think im adding anymore to this for now.

Delegus (approved):
g2g tested in game, no random spawns on other items, charges don't consume with dual mod on claws

PR #714, opened 2025-02-26
Recipe Corrections

Bakelt:
Resolves: #713
- Removed duplicate recipes that were preventing the weapon crafts working properly (Mal, Ist, Gul, Sur recipes)
- Caster crafts now properly roll 20-30


Vingelis (approved):
LGTM

PR #720, opened 2025-02-27
Update Ama Jav Automagic

Delegus:
FIlled out all the 1s in rare since it has already been done in magicprefix and suffix for most modifiers. *Shrug*

Added group 318 for Amazon Javelins

Assigned group 318 to Amazon Javelins so they will also get replenish as intended.


Bakelt (approved):
LGTM

PR #723, opened 2025-02-27
Update Controller Stash UI for Consistency

Delegus:
Note: Removed files are duplicate assets in the wrong position in the folder layout. Direct copy of data/hd/global/ui aside from my changes.

The controller stash feels out of place next to the player inventory, severely, and this aims to resolve that problem for now.

Old:
image

New: (Ignore the extra tabs, my .d2i is expanded, they won't show normally.)
image
Expanded New: (Close Button is removed but shown here cause old SS)
image


Bakelt (approved):
LGTM

PR #724, opened 2025-02-27
DruidSupportSummon-HP-Adjust

Bakelt:
- Decreased Spirit of Barbs HP/lvl
- Increased HP/lvl for Heart of the Wolverine, Oak Sage, and all Vines for Hell only


The-Honorable (approved):
LGTM

PR #726, opened 2025-02-27
Wand lvl req auto

Delegus:
Updates the auto prefix so wands can be useable at level 1.
Updates the auto prefix so javelins can be useable at level 1.


Bakelt:
Have we done the same for javs yet?

Delegus:
Added Javelins

Vingelis (approved):
LGTM

PR #733, opened 2025-02-28
Fix Aura/OOB Stat

Delegus:
Carrion Aura updated with the right stat.

Blindsight brought in-bounds with itemstatcost.txt bits.


Bakelt (approved):
LGTM - regen is the prop, hpregen is the itemstat and aurastat uses itemstat

PR #736, opened 2025-03-01
Fixes Vengeance Tooltip

Delegus:
Tooltip was not updating properly with base and next during skill point allocation and was always just showing next.


The-Honorable (approved):
lgtm

PR #738, opened 2025-03-01
Splash Bug

Delegus:
Resolves: #737 

Updated splash skill cause of an odd interaction between it, all skills and mana costs inhereted by the splash due to a combination of various flags


Bakelt (approved):
LGTM

PR #744, opened 2025-03-02
Assassin MA mosaic mod scaling changes

zakt4n:
**Idea**

Martial arts assassins now use, as a crutch, the modifier "chance for finishing moves to not consume charges" (or Mosaic mod). It is pretty much mandatory in the current state of its skills. In order to alleviate the restrictive _need_ to use dual Mosaic runeword claws -- which is restrictive for the fact that it narrows down the weapon itemization, arguably the most important item slot of a build, to just this one runeword --, I sought to give the assassin a more generic way of scaling that modifier while at the same time nerfing Mosaic itself.

**Intention:**

- Expand the ways of getting the Mosaic mod, allowing other claws that don't directly have it to help working towards the _needed_ 100% breakpoint. 
- Make the Mosaic runeword itself just a entry point claw, that can be used but can be potentially surpassed by the rare claws and/or unique claws.

**Actual changes:**
- Claw Mastery now has a linear scaling % chance for finishing moves to not consume charges (1% per level) up to level 25. After level 25, it has diminishing returns in which the player can expect to have 73% chance to not consume charges on finishing moves at Claw Mastery level 100.
- Removed Mosaic runeword's % elemental damage modifiers, reduced +2 to Martial Arts to +1 to Martial Arts


data/global/excel/runes.txt
Delegus, line 1: This entire file is showing as it was completely reworked, even if it works in game you should reload a fresh 2.0.4 version and make the changes to Mosaic you want reviewed with AFJ Sheet and make sure the difference is contained within the rune word and not the entire file.

data/local/lng/strings/skills.json
Delegus, line 1: This entire file is showing as it was completely reworked, even if it works in game you should reload a fresh 2.0.4 version and make the changes to the string you want reviewed and make sure the difference is contained within the rune word and not the entire file. Also the synergy text is recolored due to the "YC3" still at the end of the new string.

data/global/excel/skills.txt
Delegus, line 1: Walk me through the reasoning behind this method of introducing the stat to claw mastery opposed to just putting it in passivestat4 and passivecalc4 on claw mastery.

data/global/excel/runes.txt
zakt4n: That's weird. I just forked the project and downloaded it right before this change to make them. I'm aware what the programs to be used are. I'll see what I can do and why the diff is seeing it as a completely changed file.

data/global/excel/skills.txt
zakt4n: The item stat for the Mosaic mod has some funky hardcoded issues, one of them is that coming directly from Claw Mastery doesn't work. It will display correctly, but when you attack, you will still lose the charges as if you didn't have it. Solution: give it thru another state from the charge up skillls themselves. Reason why tiger strike and fists of fire is that all the other ones require those two, so when the player wants to play a MA assassin they'll need to assign points to them and thus the passive stat will work and since the stat scales based off of Claw Mastery's level, if they don't have levels on it, they won't get the bonuses; effectively working exactly as if it was just a passive stat from Claw Mastery itself

Vingelis: Are you suggesting 50% comes from Tiger Strike and 50% from Fists of Fire?

zakt4n: No, that's the opposite of what I typed. I said it is based on the Claw Mastery skill. The fact that the item stat comes from fof and ts gets completely side stepped by the fact that they are both required skills for any MA skill using charge ups and that the passive that grants the stat calls to the Claw Mastery's stats like lvl, param7 and param8.

Delegus: Claw Mastery not applying it properly doesn't surprise me, I was just genuinely curious. Had a similar issue with charm weight, which is why it is activated by weapons and not a charstats.txt passive skill.

zakt4n: Yeah, the charm weight thing is because passive stats don't get updated by the engine unless they 1) are hardcoded to update on certain frequency or 2) the items that carry them themselves get updated on certain frequency or 3) the player does something that forces it to update (usually something like assigning skill points to a skill or reloading the character, etc.). Weapons fall into the number (2) since they are one of the slots that are considered "cosmetic" (or whatever you wanna call it), and as such have functions that call them every frame to update visuals. That means that all their stats get updated every frame (it might be a bit more than every 1 frame, I'm not sure I remember it right), which in turns means we can use them to update stats properly, The "Claw Mastery + item_charge_noconsume" issue is something I'm not entirely sure why it doesn't work, largely because it's a new stat and we have no clue how they coded that into the game. One day maybe I'll be able to reverse engineer the game/portions of it and see how they did it

data/global/excel/runes.txt
zakt4n: Uh, I just realized. The d2r-reimagined/d2r-reimagined-mod repo seems to only have two .txt files in the global/excel directory to me. Is that the case for you as well? Could that be tied to why diff understood it as a completely new file?

Delegus: Ah yeah your repo has Excel capitalized so git is merging it as if the file doesn't exist, though it is using out lower-case folder at least.

zakt4n: Weird, to me it shows in the original branch itself.

zakt4n: I'm not sure how to fix it. I'm assuming that since the master branch is itself faulty, there is no real way I can fix it(?). Because I just attempted to fix the folder structure on my end (in my local fork) and push the commit to a pull request and it remains with the same issue. I may try to completely delete my fork, create a new one and ask for another PR if this issue becomes the reason why this update isn't pushed to the master branch. If that's the case, pls tell me so I can do it. I'm not sure what else I could do

Delegus:
Sorry been testing and trying out various things.

Having the entire runes.txt and skills.json files be a full differential will not get merged, as that is too difficult to see changes/differences at that point.

- Magic, Rare, Crafted claws can now have the Mosaic modifier per update 2.0.4 which reduces the need for Mosaic, especially when crafting the end-game recipes.

- The passive % is too low considering the only modifier on Mosaic can be the no charge consume and it would still be better to wear 2 Mosaics than spec into claw mastery beyond the point that is needed for BoS/Fade
  - Even if it was 100% at 25, it's not going to be used beyond finding 2 claws with no charge consume.
  - Just isn't worth the loss in damage and synergy points in the elemental skills.

- For the item_charge_noconsume
  - Moving it to passivestat1 - 3 on claw mastery it works fine, just not in 4+
  - The other passives on the skill do work in 4+
  - Works on Dragon Talon and almost every other active skill without passives, also, lol, T.I.L.

zakt4n:
Wrt the initial part of the post, yeah, the change I'm proposing here is precisely with the changes in mind. If you are making so other claws can roll the mod, then I thought it would be a good idea to allow for other axes of scaling it. The core issue I'm trying to amend here is that precisely because other claws can roll it, it makes so any claw that _doesn't_ roll it, completely useless. While this issue of a single core mod to even consider an item slot is not itself an issue in certain slots, like one may argue for movement speed in boots for example, I think that for weapon slots, which I consider to be one of the main slots for itemization in an ARPG, shouldn't be restricted by such design. Adding the mod to Claw Mastery has the goal of democratizing the access to the modifier from skill level modifiers on the claw itself, or charms, etc. effectively broadening the access to it. It isn't supposed to be a way to just get the whole 100% from skill levels alone, but to support other sources. So instead of requiring you to only look for those two mods on claws, trashing all claws that don't immediately roll it, you can get smaller rolls on one claw for example, while you focus on skill levels from other sources. Helping you get to the 100% breakpoint that is required to play the MA charge up skills as they are today. That being said, if the numbers on it are still too weak, that could be easily tweaked. I was too afraid of making it broken (like a 1 point wonder that solves the whole stat for you) that I balanced it on the extreme cautious lower end.

Wrt the second part, that's really interesting. I had tested by putting the passive stat in the first slot and didn't work, I hadn't tried to remove the h2h restriction tho. That isn't really a proper solution because you don't want Claw Mastery to work with other non-claw weapon types, but it is interesting nonetheless that a skill that restricts weapon types to h2h would make the stat not work. The curiosity on how they coded it in only grows! Lol

Delegus:
Yeah kinda confused by the h2h thing not working myself since ya know, was testing with single and dual wield claws.

Would err on the side of 50% being achievable pretty easily, claw mastery is pretty much a 1 point only unless you are purposefully playing a MA physical build, and then you would just roll crafted/rare claws anyways for the affix.

I am currently leveling a sin in 2.0.4 and around level 40 gambled 2 magic 50% claws and will likely rock them until mosaic.

Idea is solid, just trying to think of an implementation the average player will not only understand but utilize without them coming into discord and asking why we don't just give 100%.

The-Honorable:
converted to draft since its being worked on still

PR #752, opened 2025-03-03
Splash Take Something

Delegus:
- Splash now works with physical only damage.
- Splash no longer triggers CTC effects once per active frame (25 times a second.)


The-Honorable (approved):
lgtm

PR #757, opened 2025-03-04
oksill Value Fix

Delegus:
oskill with a range are currently rolling seperate values for the display string and the actual value given. This corrects that.

This fix is not retroactive, runewords such as CTA will need to be re-created, items such as Hellhunger will need to be refound in order for the correct +skill to display.


The-Honorable (approved):
LGTM

PR #761, opened 2025-03-05
Change sorc mastery o skills

Delegus:
Removed Fire Mastery and Lightning Mastery oskills from weapons and replaced with extra-fire and extra-ltng.
1h 20-25 and 2h 35-40


Vingelis (approved):
LGTM

PR #766, opened 2025-03-07
Wrath of Heaven Fix 2

Delegus:
The oskill fire mastery was replaced with extra-fire directly and I didn't realize it already had the modifier present. Condensed them together and changed the range total from 40-45 to 30-35


The-Honorable (approved):
LGTM

PR #768, opened 2025-03-07
TC Tweaks

Delegus:
- Socket orbs are slightly less common due to much higher availability by being included on champion monsters.
- All rune bosses rune table from 0 to 2 nodrop to match countess and give a minor variance.
- All rune bosses non-TZ normalized to 10 base nodrop up from 5 so the rune table should properly trigger when farmed on /p1 no-TZ
- All other bosses TZ normalized to 5 base nodrop to match their non-TZ counterparts.
- H cows brought in line with the rest of Hell


The-Honorable (approved):
LGTM

PR #772, opened 2025-03-08
Default skill icons

Delegus:
Update default game icons to be in line with Reimagined Icon art.
image
image
image


The-Honorable (approved):
icons look good!

PR #779, opened 2025-03-11
Disable Unused Souls

Delegus:
Treehead and Snapchip souls are currently unused, they can be re-added to the TC if a use is found for them.


Vingelis (approved):
LGTM

PR #782, opened 2025-03-13
Fix Warp On Controller

Delegus:
This update sets warp to the same distance as teleport when used by a controller player.

added file controllerskillsettings.json to data/hd/global/excel
- The only difference between CASC file and this is the addition of Warp as a clone of Teleport


Vingelis (approved):
LGTM

PR #784, opened 2025-03-14
foth desc update

The-Honorable:
- updated fist of the heavens description to properly display cast delay


The-Honorable:
canceled del has same one up

PR #785, opened 2025-03-15
Tooltips and Timings

Delegus:
Updates CD tooltips for FORB Firewall and FOTH to the right timings. 
Updates skills.txt for the right timings.
Update Vengeance tooltip so AR% shows properly.


Vingelis (approved):
LGTM

PR #786, opened 2025-03-18
Grizzly Sunder Fix

Delegus:
Summon Grizzly Bear now inherits physical sunder as it should.

Unless this was intentionally left off, in which case this can be closed.


The-Honorable (approved):
LGTM

PR #797, opened 2025-03-25
Dye System Expansion

Vingelis:
Unlinking issue so it remains open.


Delegus:
At this point this will just need to be remade as it is very out of date with the current state of the base mod. I'll unlink the issue.

PR #800, opened 2025-03-28
Pet icon trap count Tec Slam Tooltip

Delegus:
Adds portrait for traps to show the # spawned. Updates on summon and end of life (shot charges depleted)

image
image

Note: Due to how traps are all a single pettype "assassintrap" this prevents more than 5 traps from ever being summoned regardless of trap type. 
- To display separate icons per type of trap this would need to be split up.
  - Splitting them up will remove this limit without a lot of formula work in the skills.txt file and isn't worth the effort unless we want to just allow 5 of each type of trap at a time.

Updated Tectonic Slam damage tooltip to properly display damage when allocating points.


The-Honorable (approved):
LGTM

PR #805, opened 2025-04-01
Remove Summon Armor

Delegus:
Armor is a placeholder that is no longer used and causes players using D2RMM and Psuedo LF to crash because it causes it to spawn in A1 N lower level zones.


The-Honorable (approved):
lgtm game launches/joins.

PR #810, opened 2025-04-10
tweaking poison/plague jav

The-Honorable:
- Casting delay removed
- Duration set to 2 seconds at all levels
- Synergy upped from 12% to 15%
- Damage curves reworked at all levels
  - Heavily buffed damage curve past 29+
- Poison trail velocity increased.
  - Trails moves much faster behind the thrown javelin
- Poison Trail duration reduced 

- Removed the ability for Plague Javelin to pierce
- Plague Cloud expand velocity increased
- Duration starts at 3 and scales to a max of 6 seconds at level 16
- Adjusted synergy from 15% to 10% to compensate for the lowered duration
- Skill description updated to mention the javelin does not pierce

- Particle effect on the flying scimitars adjusted to cover the whole sword.


Delegus (approved):
Time for merge conflicts

PR #814, opened 2025-04-11
Arctic Claw in Tree

GildyBoye:
- Arctic Claw in Tree, Maul swapped out
- Feral Rage can be used in all shapeshift forms
- New unique to obtain polar bear at endgame outside of Set

image
image


Delegus (changes requested):
This PR is a combination of all changes in this and the prior PR please separate them out or close 1 or the other.

PR #817, opened 2025-04-13
Initial Dump

Bakelt:
- New CtC's for Clay Spire, Thunderbolts, Spiral Tornado, Frost Breath, Holy Orb, Throwing Axe
- Cloned skills for CtC use for Nova, Frost Nova

To Do:
- Create skill curves
- Fix dmg bug on Clay Spire
- Clone more common CtC's


Bakelt:
Ideally will replace Holy Orb with a teeth version if I feel like getting around to doing the texture replacements for teeth so it doesn't originate from the player.

PR #823, opened 2025-04-17
Assassin tinkering

Delegus:
Stopping with these changes on this Assassin PR due to plans Uthstar has for the Shadow Disciplines in the future.

Blade Fury Reworked
- Now throws a shuriken using attack speed that bounces to up to 2 nearby enemies (2 bounce limit doesn't scale)
- Gains 1 projectile per 8 soft skill levels. (12 at skill level 100 but may hard cap with a min formula pending tests)
- Carries weapon damage and modifiers

Blade Sentinel
- Now does 100% weapon damage
- Curves slightly increased

Blade Shield
- Curves slightly increased


The-Honorable (approved):
LGTM

PR #831, opened 2025-04-23
Fix Skill ID Comments

Delegus:
Comment Column updated to reflect actual ID used by the game engine.


The-Honorable (approved):
lgtm y33t it into the void

PR #832, opened 2025-04-23
Laggy Missiles

Delegus:
Vines cause lag, removed scaling duration because it doesn't do anything useful and contributes to lasting effects which greatly tank the games FPS.

War cry echoes can cause crashing. Reduced the number of echo missiles spawned per war cry nova by half, just extra visual with no purpose due to the NHD on the skill.

Battle Cry visual updated and fixed so that it shows the correct range when cast.

Skill description updated for range of Battle Cry

Howl and War Cry damage curves increased 29+ to improve hell gameplay and overall feel of the skills late game. Synergies both to 15 from 12 and 10.


The-Honorable (approved):
LGTM

PR #837, opened 2025-04-26
Add % Attribute to Properties

Delegus:
str%, dex%, vit%, and enr% added as working affixes to itemstatcost.txt to compliment existing properties.txt entries
Stat Range: -63 to +64% attribute bonus per individual item.


The-Honorable (approved):
LGTM % affixes work on the charm tests

PR #838, opened 2025-04-26
Fix Ancients Bug

Delegus:
Re-flag ancients with stack so that they can spawn more than one time in a single game and stop playing DDR on their pedestals.


The-Honorable (approved):
LGTM

PR #840, opened 2025-04-28
bone armor pdr to def

The-Honorable:
- Physical damage reduction per level removed
- Added flat defense per level. 50+50/lvl


Delegus:
You may not pass. - A2 Guard Whose Name I Forgot 🤣

Delegus (approved):

PR #841, opened 2025-04-29
Anni Upgrade Path Revision

### LINKED ISSUE #698: Torch/Anni upgrade refiniment
The new addition of an upgrade to these two, while at it's core a very nice idea, kinda feels unrefined and inconsistent right now.
The Torch upgrade requires you to clear Ubers a total of 11 times (Base + Upgrades) while the Anni only needs 1 clear and instead requires 1 of each Orb (except Socketing) to be cubed with it. Also you are able to stash those 10 additional torches in your cube for other characters to use during this process, the same cannot be said for the Anni as you will be spending the Standard of Heroes and are therefore "sacrificing" an Anni.
Yes, they don't need to be identical in how they are eventually upgraded but it is very clear that the time and effort needed between those two differs significantly. I'd even argue the former process is a bit too tedious and boring, altho understandable to somehow spent the Keys you are passively getting from farming.
Here is my solution:
- Reduce the upgrades needed for the Torch from 10 down to 5 (2 to stats per upgrade instead of 1)
- Implement a similar system to the Anni and make it require 5 cube upgrades as well (whether that means changing the recipe slightly or increasing the drop rates due to specific Orbs being very very rare AND needed in other recipes constantly would be up to you)
- The Torch requires 1 more column in the inventory after upgrading, yet the Anni does not despite gaining "Teleport" as an oskill. 
Therefore, perhaps adding something like "Warmth" as an oskill to the Obsidian Beacon could be worth looking into to warrant the effort, more space used and overall fit thematically/be consistent with the Anni. Should be a somewhat useful skill passively to get around Mana issues without having to full send it on Leech or per kill suffixes.

Vingelis:
Fixes #698 

Anni Upgrade path to Black Soulstone revised

Black Soulstone: 2 ASK, 20 Stats, 20 Res, 10 XP, 5 Teleport (down from 2 ASK, 25 Stats, 25 Res, 10 XP, 5 Teleport)

Annihilus upgrade path now has 6 levels. Each upgrade level adds stats, similar to torch. Reagent costs reflect Orb drops per act that are being added in TC changes (to keep things thematic).

First, JAH + ITH + BER = + Lvl 5 Warp OSkill
Second, 3 Infusion Orbs = +1 XP, +2 All Stats, +2 All Res
Third, 3 Assemblage Orbs = +1 XP, +2 All Stats, +2 All Res
Fourth, 3 Corruption Orbs = +1 XP, +2 All Stats, +2 All Res
Fifth, 3 Conversion Orbs = +1 XP, +2 All Stats, +2 All Res
Final, 3 Shadow Orbs = Black Soulstone

<details>
<summary>Upgrade testing screenshots</summary>

- base anni

image

- First, JAH + ITH + BER = + Lvl 5 Warp OSkill

image

- Second, 3 Infusion Orbs = +1 XP, +2 All Stats, +2 All Res

image

- Third, 3 Assemblage Orbs = +1 XP, +2 All Stats, +2 All Res

image

- Fourth, 3 Corruption Orbs = +1 XP, +2 All Stats, +2 All Res

image

- Fifth, 3 Conversion Orbs = +1 XP, +2 All Stats, +2 All Res

image

- Final, 3 Shadow Orbs = Black Soulstone

image

</details>


data/global/excel/uniqueitems.txt
```diff
@@ -1403,7 +1403,7 @@ Quiver of Amplified Slaying	1429	100	1			1	1	1	85	z01	Arrows		5	5000	blac	blac
 Quiver of Resistance Slaying	1430	100	1			1	1	1	85	z01	Arrows		5	5000	blac	blac		invarro3				dmg-norm		10	25	swing1		20	20	pierce		25	25	skilltab	0	2	2	rep-quant	100			hit-skill	91	20	5	oskill_hide	393	1	1																						0
 Bolt Case of Amplified Slaying	1431	100	1			1	1	1	85	z02	Bolts		5	5000	blac	blac						dmg-norm		10	25	swing1		20	20	pierce		25	25	skilltab	0	2	2	rep-quant	100			hit-skill	66	20	10	oskill_hide	393	1	1																						0
 Bolt Case of Resistance Slaying	1432	100	1			1	1	1	85	z02	Bolts		5	5000	blac	blac						dmg-norm		10	25	swing1		20	20	pierce		25	25	skilltab	0	2	2	rep-quant	100			hit-skill	91	20	5	oskill_hide	393	1	1																						0
-Black Soulstone	1433	100	0			1	1	85	85	cm1	charm		3	5000			flpmss	invmss	item_gem	12	item_gem	allskills		2	2	all-stats		25	25	res-all		25	25	oskill	Teleport	5	5	addxp		10	10	charm-property		1	1	uberstone-property		1	1																						0
+Black Soulstone	1433	100	0			1	1	85	85	cm1	charm		3	5000			flpmss	invmss	item_gem	12	item_gem	allskills		2	2	all-stats		10	10	res-all		10	10	oskill	Teleport	5	5	addxp		10	10	charm-property		1	1	uberstone-property		1	1																						0
```
Delegus, line 1406: Considering how massive of a resource sink this is, it feels bad to be half the stats of vanilla Annihilus but +2 ASK and +5 Teleport do make up some of the difference. Personally not a fan of the stats/res being less than 20 on the final charm considering the time and effort it takes to farm it with this pathing update but the others can weigh in here.

The-Honorable: think i agree here. even just leaving it to 20 or a variable 15-20

I was also curios why 5 tp instead of just 1

Harald92: I am on the same boat of 15-20 on the stats. The teleport does not bother me much, the only benefit is lower mana cost, which could help some low mana builds.

The-Honorable:
converted to draft while i sort out the upgrade path problems

The-Honorable:
upgrade path fixed

Delegus (approved):
still taking passengers

PR #843, opened 2025-05-01
swap skill rand skill for an actual number

The-Honorable:
### Reason
- Random Sorc skill will roll at +4 now

### Terror
- Random Necromancer skill will roll at +4 now

## Faith's Promise
- Random Paladin skill will roll +5 now


Harald92:
Looks good.

I checked the other files really quick and other than that there are some uniques which gain 8-10 skills, which is not possible to my knowledge, 7 is the max, I have not found any other instanced of this typo.

Harald92 (approved):
All well

The-Honorable:
updated to include faith's promise unique

Harald92 (approved):
All done :)

PR #844, opened 2025-05-02
CTC Integration

Harald92:
### **CTC Integration**

- The goal of this PR is to enable skills of every Class to be usable as a CTC.
- Only working skills got enabled, not working or bad working ones did not make it to be save.
- The change itself does nothing to the game right now, it just enables the changed skillt to be able to used for CTC in the future.

**The Process:**

- I tested each skill in game and checked how and if they work as CTC when enabled. Some skills work without problem, some do work with limitations, some do not work at all. Some skills need some work to be done to be useful/usable as a CTC skill:
1. Damage skills need an improved/fix damage curve to be of use of classes without the possible synergy boni. Achievable is that by creating a copy of those skills as CTC skills.
2. Some skills only work when you have skills inside the same or other skills. Achievable is that by creating a copy of those skills and apply them as "oskills" but making them not usable with left and right click.
3. Some melee skills only activate as gethit skills when the player is attacking, they work but have that limitation.
4. Some missile skills spawn their projectile at the target location, this can look funky.
5. Some missile skills use up ammunition, which can cause problems.

**Documentation:**

- I documented my findings [here](https://docs.google.com/spreadsheets/d/1HbcMTqFxatbDgpHHHTSwX4q1kXz3Aq8bc0DXi6FKrN4/edit?gid=903107061#gid=903107061). You can comment if you have something to add! (The legend for the colors of the skill names in the top left as a comment in the cell C3 names skills)


Delegus:
Will review once conflict fix is done.

PR #849, opened 2025-05-09
Panda Pots Duration Stacking

Delegus:
Update the panda potions so that the duration stacks similar to thawing/antidote potions.

These aren't seeing use because of the price and it is unwieldy to keep track of their duration and even with the visual effects it is not always obvious mid battle when they fall off. Changing them so that the duration can stack.


The-Honorable (approved):
stacking works. LGTM

PR #851, opened 2025-05-11
Warp Tooltip

Delegus:
Since we are moving towards adding warp to more items I have updated the tooltip to reflect the cooldown and updated mana cost so it doesn't bypass charm weight.


The-Honorable (approved):
lgtm

PR #855, opened 2025-05-18
Fix of the Swordsman

Delegus:
fix suffix of the swordsman is missing a string entry.

fix montype path casing

fix mind blast tooltip from 6 - 0 to 6

fix death sentry corpse pop to properly show 40-80 instead of 40-0


The-Honorable (approved):
lgtm

PR #856, opened 2025-05-18
version update

Delegus:
update version text to 2.1.0, this same element is pulled by controller ui so doesn't need updating there.


The-Honorable (approved):
lgtm

PR #862, opened 2025-05-23
Barb Throwing Stuff

Delegus:
# Barbarian Skills

Updated Shield Mastery and Deadly Teasures Icons to use the correct border style in keeping with all other icons and reworked both slightly to make a bit more sense. You can see shield mastery below.
image

Changed "Combat Masteries" to Masteries and Throwing Skills

All 4 new skills synergize with each other.
image

### Unstoppable Force (Replaces Pole-Arm Mastery)
Level 1
- High Damage Slow Moving Projectile that pierces enemies and always hits
  - AR Based alternatives felt horridable

### Winters Gambit (Replaces Spear-Mastery)
Level 12
- Throws multiple projectiles that travel for a short distance ignoring collision and then home in on random "nearby" enemies and zips to them to deal weapon damage and cold damage
  - Skill that engages you and rewards positioning, don't hug the monster unless it is alone if you want to hit it. 🤣 
    - Inspired by Storm Crows from Median by Brother Laz

### Chasm Break (Replaces Increased Stamina)
Level 24
- Throw a weapon that  appears to teleport into an enemy and causes of rocks to fall on them for a short period dealing periodic damage to enemies nearby the target
  - Credit to ESR celestialray for the missiles.txt entry and Median by Brother Laz for the idea from Rock Shower

### Madawcs Fury (Replaces Grim Ward)
Level 30
- Enhances player throwing damage
- Adds flat cold damage to attacks
- Enhances AR
- Gives you a cool glow created for the skill.

# General Items
- Fixed All Max Res Property Group
- Fixed Player Damage % Property Group
- Added Player Throwing Damage % Property and Group
- Removed dmg-throw and replaced with dmg-norm (Not Grief Type) on items
  - This is due to not having a string, not being able to display as a combinable string, not having a reason not to be flat damage in general, and docgen doesn't handle the property.


The-Honorable (approved):
lgtm

PR #863, opened 2025-05-24
Fix Decoy Perma-Beam

Delegus:
So after some tracking of the beam itself it is added to all the player model corpses to make finding them easier.

The decoy is a duplicate of the amazon and thus also received said beam on death.

Flagging the state as hidedead solves the problem.


The-Honorable (approved):
lgtm

PR #864, opened 2025-05-24
MS Sound Bug/Ice Blurgsh

Delegus:
Re-added sound compounding delay to both multiple shot missiles and thankfully, thankfully, it no longer plays at 4x volume while shooting.

Also added 3D spread back to ice shatter and punch because they both go too hard when standing close.


The-Honorable (approved):
lgtm

PR #866, opened 2025-05-24
Update modinfo.json

Jasper89c:
Added version code to modinfo for launcher


Jasper89c:
This will need to be updated for each patch before release

CollinHerber (approved):

PR #867, opened 2025-05-24
Deadly treasures

Delegus:
Fixed maximum stat overflow

Gains 4 % per level MF and 6.6% GF
- Still a 1 point wonder but not as absurdly as before considering it would reach 1023 and 2047 now if I had left the scaling alone.


The-Honorable (approved):
lgtm

PR #869, opened 2025-05-28
MS GA Synergy Fix

Delegus:
Accidently set it too low by 1 to maintain damage parity with 2.0.5

When updating patch notes, reword to specify 25 * 12 * 0.75 = 25 * 9 in damage.


The-Honorable (approved):
lgtm

PR #870, opened 2025-05-30
Rune Rebuild

Delegus:
Ignore Orbs, runes are the only change in this PR

image

Example of a random assortment from a playthrough character.

image

Rebuilt Rune Sprites for visual clarity
- Stack icon made brighter
- Runes are less visually noisy, used base game sprites as a template instead of the ones with the sweeping bevels
- Colors kept and kept on the same numbers

Orbs are a no for now in this PR.

Fixed inventory grid positioning being off by a pixel after using a testing sprite to verify that I had indeed misaligned them all by 1 before.


The-Honorable (approved):
lgtm

PR #887, opened 2025-06-12
Set Overhaul - Class specific sets

Harald92:
[Set Overview Spreadsheet](https://docs.google.com/spreadsheets/d/1NMku2jKMteQs-Md6KQIvWKb2qI-h9Mu4WMZwtstiPOI/edit?gid=787945896#gid=787945896)

[Safe files for each class for testing](https://drive.google.com/drive/folders/1qUFpiMR--xpt9vkKpK17Xpx8aLJ8yMsm?usp=sharing)

**The Goal of the Overhaul of sets**
- Create and/or change existing sets to make 6 sets for each class for different level stages which operate on specific rules for each stage. Stage rules for those sets can be seen in the spreadsheet's sheet called `Info`. I have chosen sets to use as the class sets which were very near stat and level wise to my rules already, those can be seen on the `Set Overview` and are colored like the state rule theme stated in the `Info` sheet.
- Change existing sets, which are not used as a class set, to meet rules which set needs to accomplish. These rules are stated in the `Info` sheet too.

**Progress**
- In the `Set Overview` sheet if there is a checkmark in column **C** beside the set name this means I have overhauled the set. By using the hyperlink in the set name cell you can see the changes I made. The changes I made are vast and it makes no sense to state everything I plan to change here.
- The stats I choose for sets and set items are compared to already existing set items, unique items and runewords.

**Comments and suggestions**
- If you have suggestions just comment in the spreadsheet, leave a comment here or PM me on Discord.


The-Honorable:
I noticed a few functionality issues/conflicts like items having raven stuff and shift forms stuff. you can cast ravens while shifted so these make no sense.

The-Honorable:
as a whole i agree with del. i dont really think the sets here are doing anything inspiring and seem very cookie cut/bland. which is fine for some, but not every set doing basically the same thing for x element on y class. especially the same oskills on everything. 

honestly id rather see sets reworked in a way that alters builds or provides unique way to play a class/build. kind of like how raven set removes max wolves but buffs ravens as your main damage source instead of ravens just augmenting summoner/not being used that much.

Delegus:
Not 3.0 updated, no response to feedback from over 9 months ago.. Closing.

PR #898, opened 2025-07-06
BF Tweak

Delegus:
Adjusted skill so it bounces from enemy to enemy or from enemy to wall to enemy instead of just splitting from 1 to 2 to 4 like Lightning Fury


The-Honorable (approved):
lgtm

PR #903, opened 2025-07-13
Replace runeword id of blood for translation

d3m3vilurr:
I don't know reason, original id is mapped to `item-nameaffixes` version.

that results, translation result is not correct.
original translation and users expected `피`, in game message is `피의` (of blood)

this patch replace runeword id to `Blood1` from `Blood`, after a patch, game present item message correctly.


d3m3vilurr:
I don't know that this is correct patch approach.
and also, I think other some runewords(bone, bramble, ..) would have same problem.
but in this time, I only checked blood which reported bug case.

d3m3vilurr:
I checked that `bramble` has same problem, translation data expect `찔레` but present `찔레의`.
it's same problem of the `blood`.

I'll make a script to check conflict id for fully fixing this case.
but anyone confirm this fixing approach is acceptable or not?

Delegus:
Yeah changing to Blood1 for the key is fine, probably a mix up back when I fixed around 2K missing string problems due to numbers in the txt files but not in the keys. I did the same for the Arcane runeword Dys added, made it Arcane1

d3m3vilurr:
checked 4~5 runewords, works well :)

Delegus (approved):

PR #905, opened 2025-07-18
Initial Commit

Dracarys33:
Initial update of Obsidion mod to sync changes to Github


The-Honorable:
there are a lot of questionable things here we likely dont want in the mod or want to scrutinize heavily.

PR #913, opened 2025-07-23
Splinterfreeze CTC Update

Delegus:
Clone frost nova to new CTC skill 100-150 damage per skill level
Skill cannot be used manually even when set as an oskill (intentional)
Currently only used on splinterfreeze as a level 1 CTC because item is level 32 required.


The-Honorable (approved):
lgtm

PR #917, opened 2025-07-26
Rando Fixes

Delegus:
Fix Swift Descent Spelling
Fix Tyrial's > Tyrael's Spelling
Fix CB Sentry Synergy Tooltip
Update TZ Rota to remove Stony/Trist from A1 and add it to A4, added Tamoe and Outer Cloister to A1 (At 24 zone limit)


The-Honorable (approved):
lgtm

PR #945, opened 2025-08-09
Fix Misc Issues

Delegus:
Fixes broken splash charm recipe
Fixes broken CTC Frost Nova that got booted out in a merge somewhere.


The-Honorable (approved):
lgtm

PR #951, opened 2025-08-10
update cow portal recipe

The-Honorable:
# Recipes
> Adding back an optional vanilla cow portal recipe since many users fail to read or realize the recipe was updated to just leg

- Vanilla cow portal recipe added back
- Reimagined cow portal recipe is still just Wirt's leg
- Reimagined cow portal recipe now returned Wirt's leg after opening the cow portal
  - You still need the correct wirt's leg for each difficulty


Delegus (approved):
Moist moos moistly mooing

PR #954, opened 2025-08-11
Carnage Tweak

Delegus:
Carnage Visual and Feel Updates
- Tornado is still visual only and spawns on swing
- Spikes spawn on weapon swing instead of after tornado animation finishes
  - Spikes updated visually with red glow to make it much more apparent what the skill is doing.
- Swing is full weapon damage
- Spikes are still half weapon damage

Improved overall feel of the skill as there is a much more immediate impact instead of the 1.5-2s delay from swing to spike re-target and damage dealing, updated visuals give a much better idea of what the skill is doing and where the damage is coming from and it just looks even cooler.


The-Honorable (approved):
LGTM awesome changes

PR #955, opened 2025-08-11
thorns works on ranged now

The-Honorable:
# Thorns itemstat
- updated thorns itemstat so it works against ranged attacks

# Paladin
## Thorns Aura
- Increased the amount of flat thorns given by the aura, mainly the e5 so 29+ curve is very big
- Also increased the flat amount give by 1% per character level

examples on a level 99 character:

image

image


Delegus (approved):
watery wet moistness

PR #957, opened 2025-08-13
Kings Trinity & IK Set Updates & Vengeance

### LINKED ISSUE #961: Barb Howl Close Cast Damage Failing
Howl is bugging out at close range and not dealing damage (same issue I was having while reworking shockwave, caused by a bad svrdo function)

Delegus:
Resolves #961 

Howl
Removed initiate frame since missile is so fast it causes it to miss initial target sometimes.

Kings Trinity
Oskill that does a physical hit and spawns 3 projectiles that circle with a few frames of delay between them adding an additional 50% of weapon damage as fire, cold and lightning.

Vengeance
Same as above but 25% weapon damage with state buff synergy weapon % scaling.

Initial hit carries all weapon effects.
Spawned projectiles do not.

# Sets

- On full set receive +1 King's Trinity Oskill
> King's Trinity create cold/fire/lightning tornado nova on strike. Initial strike is physical, and each elemental missile is converted from 37% of weapon damage to that element. 

- On full set receive 10% Chance to cast level 1 King's Charisma
> King's Charisma provides a flat +1000-2000 cold/fire/lightning damage to attacks

### immortal King's Will
- Remove socket rolls
- Now rolls +2 random barbarian skill
- Now has a 2 set bonus of +15% Faster Hit Recovery
- Now has a 4 set bonus of +2 barbarian skills

### Immortal King's Soul Cage
- Removed chance to cast level 5 enchant when struck
- Added an enchant clone with its own unique skill curve: 10% chance to cast level 1 CTC Enchant when struck
  - level 1 CTC Enchant provides 500-1000 Flat fire damage and 100% increase attack rating

### Immortal King's Detail
- Added 10-20 Weapon Damage
- Added +2 random barbarian skill

### Immortal King's Forge
- Added +2 random barbarian skill
- Removed chance to cast charged bolt when struck
- Added a charged bolt clone with its own unique skill curve: 20% chance to cast level 3 CTC Charged Bolt
> CTC Charged Bolt has a level scaling of 50-100 per level and 5 bolts + 1/lvl

### Immortal King's Pillar
- Added +2 random barbarian skill

### Immortal King's Stone Crusher
- Removed indestructible
- Added Ethereal
- Crushing blow reduced to 15-25% from 35-40%
- Increased Enhanced damage from 200% to a range of 200-300%
- Flat damage given for each set bonus stage increased


The-Honorable (approved):
lgtm

PR #965, opened 2025-08-19
Cube QoL

Delegus:
Leech nerf backed up by a good chunk, divisor to 8 from 12, no point in killing mana leech so hard, so it is also backed up.

Nightmare res penalty to -60 from -40 so it's half of Hell Penalty 

Conviction
- Retuned so that it's base 30 + 1 per level until 25, then it gains 1 per 2 levels after.
- Base points increase - resistance bonus by 5% ish per point. 
- Defence reduction spread from 0-100 instead of 40-100
>Been meaning to recurve this skill, here is as good as anywhere else considering the massive amounts of changes to skills.txt due to the gamble fix anyways.

Cube
- Uber Unique Stat Reroll Recipe now works with Set items as well

Potions
- XP, MF, Sunder potions increased to 2 minutes up from 30 seconds

Charm Weight
- Fix automagic, skills properties, and the armor and weapons so that the weight system no longer applies a minimum level requirement of 7 to items.
>This also solves the problem with level 6 spells useable at level 1 and 12 useable at 6.

Skill gold cost affect
- Skills no longer have much effect on the cost, sell price, or repair price of items, with the exception of Teleport, which will still be substantial.
>This solves infinite gambling and gives incentive to use charged items more by not being so expensive to repair, with the exception of teleport. which will still be substantial.
- Teleport is unaffected by this change, it will still cost a substantial amount.


The-Honorable (approved):
lgtm

PR #973, opened 2025-09-25
BK rework initial thought

The-Honorable:
maybe adjust the req level a bit upwards

+3 rand barb skill on each

image

image

image


The-Honorable:
think im done with my initial thought. atm there is only 1 set weapon for these bases. could adjust in the future, or make them both colossus blades since only barb can dual wield 2h swords anyways and orbing means you get 1 of either instead of guaranteed both.

benjimon24:
i love the idea of bumping up original sets like BK, but i caution against making the stats too "generic". +skills. IAS, ED, CB, DS, AR, +demons, +undead, +all res, +all stats - it just has ALL the good stuff, and makes it easy to build around. the only unique thing about it is CTC amp damage, which is another great attribute to have, but it doesn't really give it much of an identity as it just further increases physical damage

i think we should try and keep unique stats like the +fire res and +poison damage of BK's tribal guardian in some form. if we're aiming for end game viability i think it would be interesting to explore some sort of CTC poison skill, or giving it max fire res or fire absorb instead of basic fire res. 

in my opinion, BK being the original "dual wielding set" - i think i would like to see stats tailored towards that playstyle, with points towards frenzy, double swing, or melee mastery instead of being random barb skills.

i think dual wielding naturally pairs with CTC abilities, as one of its core advantages is a fast attack speed, so id love for it to have something cooler than amp damage, where barb already has war cry as a source of breaking phys immunity - maybe shockwave or some poison skill as mentioned earlier

on a side note, i also think it might be a uniquely cool idea to give the CB the paladin's charge as an oskill, given that its called "Bul-Kathos's Sacred Charge"

i also highly object to the idea of making them both colossus blades, as one big sword and one smaller sword feels very much tied to the identity of this set

anyways, just my random thoughts take it as you will

The-Honorable:
> i love the idea of bumping up original sets like BK, but i caution against making the stats too "generic". +skills. IAS, ED, CB, DS, AR, +demons, +undead, +all res, +all stats - it just has ALL the good stuff, and makes it easy to build around. the only unique thing about it is CTC amp damage, which is another great attribute to have, but it doesn't really give it much of an identity as it just further increases physical damage

- at a high level the stats the set granted were just increased across the board. it was fairly generic to begin with acting as a mid level dual wield set. better in vanilla due to lower power scaling. now it can function into hell somewhat well until you decide to craft or unique hunt and burn some orbs/upgrade mats.


> 
> i think we should try and keep unique stats like the +fire res and +poison damage of BK's tribal guardian in some form. if we're aiming for end game viability i think it would be interesting to explore some sort of CTC poison skill, or giving it max fire res or fire absorb instead of basic fire res.

- I do think i may alter they a bit to be slightly more unique ish. however, the poison damage may have just not existed at all. it basically functioned as an open wounds rather than doing any actual damage. that being said, id rather just boost the damage of the items and yeet the useless stat. i dont really see the fire res being a real benefit on a weapon and instead i turned str into all-stats as a compensator for removing the res-fire. it loses a bit of some flavor i agree, but the overall usefulness is increased.

> 
> in my opinion, BK being the original "dual wielding set" - i think i would like to see stats tailored towards that playstyle, with points towards frenzy, double swing, or melee mastery instead of being random barb skills.

- the items were each given a +1 all barb skills. while you could certainly just add +3 to frenzy or any other DW skill, there are other skills that are useful to have +3 for. battle cry/bo for example, madawcs also works for any melee skill. with the changes to some barb skills you will likely want some other +skills than just combat skills too.  it can certainly roll frenzy/double swing still. as it stands these are 100% chance to orb and would be significantly too easy to roll into if they granted specific +x to a skill imo.

> 
> i think dual wielding naturally pairs with CTC abilities, as one of its core advantages is a fast attack speed, so id love for it to have something cooler than amp damage, where barb already has war cry as a source of breaking phys immunity - maybe shockwave or some poison skill as mentioned earlier

- i am somewhat in favor of adding more flair, the idea behind amp being - enemy PDR is a straight up multiplicative damage increase on top of battle cry. with there being few ways to actually lower enemy PDR amp is great. some unique ctc would be cool though. i could add dust devils a new ctc we created but that exists in a separate PR currently. There are plans to add some new things to Frenzy so i'm hesitant to do anything too crazy to the set than just increase its overall viability atm.

> 
> on a side note, i also think it might be a uniquely cool idea to give the CB the paladin's charge as an oskill, given that its called "Bul-Kathos's Sacred Charge"

- this is an interesting idea but barb does not have the animations to charge. havent dug into adding animations before but if i do id circle back to the idea, thanks.

> 
> i also highly object to the idea of making them both colossus blades, as one big sword and one smaller sword feels very much tied to the identity of this set

- likely wont change it, but future state would be ideal to have other items on the base. depending on if dys or someone else circles around to updating sets that is likely not a 2.2 thing.

> 
> anyways, just my random thoughts take it as you will

benjimon24:
> This is `inherently wrong` and it's definitely coming from a vanilla way of thinking. `Nothing wrong about that.`

this line is just very confusing. 

>Barb already had a poison skill called Whirlwind and it deals just about damage as Poison does.

no idea what you mean by this

my main argument is that we shouldn't just give everything all stats, all res, etc. items should each have their own unique flavor instead of varying degrees of the same generic pool of "good stats". while the fire damage and poison damage is pretty insignificant in this case, its just something that sets it apart from everything else - its "flavor". id like to see these unique properties emphasized. in the end, i want to end up in a position where i think "okay, BK may not be best in slot but it has a particular unique property that is just so damn cool!" for me, CTC amp damage isnt exciting enough to make me choose it over a BIS.

on a tangent - we could also draw inspiration from D3, as the BK set exists in that game as well - if i remember correctly, it distinguishes itself there as the speed run build, offering a high degree of extra movement speed compared to other options. however, im just throwing it out there - i dont have a strong opinion of what BK's identity should be

lastly, i think its extremely important not to make resistance so easy to cap out. resistances generally isnt an issue for barbs since they have the natural resistance skill but in general capping out resistances should be difficult and comes with tradeoffs. having an item ONLY provide a high degree of fire res might force you to choose other items or charms to make up for the other resistances elsewhere. this makes itemization difficult and interesting.

The-Honorable:
> > This is `inherently wrong` and it's definitely coming from a vanilla way of thinking. `Nothing wrong about that.`
> 
> this line is just very confusing.
> 
> > Barb already had a poison skill called Whirlwind and it deals just about damage as Poison does.
> 
> no idea what you mean by this
> 
> my main argument is that we shouldn't just give everything all stats, all res, etc. items should each have their own unique flavor instead of varying degrees of the same generic pool of "good stats". while the fire damage and poison damage is pretty insignificant in this case, its just something that sets it apart from everything else - its "flavor". id like to see these unique properties emphasized. in the end, i want to end up in a position where i think "okay, BK may not be best in slot but it has a particular unique property that is just so damn cool!" for me, CTC amp damage isnt exciting enough to make me choose it over a BIS.
> 
> on a tangent - we could also draw inspiration from D3, as the BK set exists in that game as well - if i remember correctly, it distinguishes itself there as the speed run build, offering a high degree of extra movement speed compared to other options. however, im just throwing it out there - i dont have a strong opinion of what BK's identity should be
> 
> lastly, i think its extremely important not to make resistance so easy to cap out. resistances generally isnt an issue for barbs since they have the natural resistance skill but in general capping out resistances should be difficult and comes with tradeoffs. having an item ONLY provide a high degree of fire res might force you to choose other items or charms to make up for the other resistances elsewhere. this makes itemization difficult and interesting.

piggy backing on the last comment, 

I get what youre saying and i agree to a degree with the unique-ness aspect. that being said, the all res is already on the set in vanilla and currently in the mod(im not adding it). in fact, im actually removing total res from the set by removing the fire res from guardian sword.

The set bonus giving what it does is already there(demon dmg/undead dmg), im just increasing it and removing the flat fire for flat phys because its relatively useless to any barb as they scale phys and no amount of flat fire on the swords would make it desireable for a phys barb with how stat caps work.  Also removing crushing blow from one sword and moving it to the set bonus because that is a local stat and only affects attacks made by the sword that has the stat.

as for the d3 set aspect, the set could do something like increase how much ias/move speed you get from frenzy but then its pigeon holed into being a frenzy set instead of working well for any dual wield type set. im more inclined to go back and drum up something unique for the set. amp damage being useful for any melee barb and even for a5 mercs. adding one of the new CTC only skills or making some new unique ctc like skill for the set. dunno yet.

The-Honorable:
adding bk changes to pr922 with the new ctcs

PR #984, opened 2025-11-11
Animation Investigation

Delegus:
- Fixed Gem Bag sprite so the placement is more in line with the inventory grid now that it's all squared up.
>Investigated Jab Animations due to Oskill and Charges and determined the skill doesn't work with other classes because they don't have the correct sequence information. Using the amazon information results in bad TPoses, leg wobble and wiggle, and the weapon spinning in their hand, lol.
- Replaced Jab oskill on Jealousy runeword with Impale
- Removed Jab charge spawn on non-amazon items.
>Per popular demand and abundant sources of oskills for them, Zeal and Whirlwind.
- Added animation references to all character classes for all weapon types for Zeal
- Added animation references to all character classes for all weapon types for Whirlwind
  - Sorceress orbs are not flagged as a melee in itemtypes so they still aren't useable by WW but work with it if they are enabled later.
- Whirlwind
  - Damage scaling per level increased from 5 to 10%
  - Synergy increased to 5 from 4%.
>Numbers buff for Whirlwind for now, as the skill can still work quite well when used properly. It is a point to point click skill, not a hold down the button skill like in modern games and adjusting that functionality is beyond current mod tools.

>Paladin charge animation restored to vanilla function.
- The kick was unintuitive and confusing players who play both mod and vanilla and didn't match hit sounds.
>Assassin Dragon Talon animation restored to vanilla function.
- Testing revealed no significant reason this was changed. OG Talon does not have annoying noise per kick like animation change does.


The-Honorable (approved):
lgtm

Issue #681, opened 2025-02-21
Knockback suffix on Melee Weapons / Improvements


FWCrash:
Knockback as a stat for 99% of cases is a hinderance, especially for melee classes cuz they are by nature slower in clearing areas than casters. Therefore, the stat increases the chance to clear enemies even slower significantly. KB should be a choice someone can make, cuz it is definetly warranted for clearing single target uniques, but overall not so good. The only exception to this are Bowazons and perhaps an A2/A5 mercenary. 
Solution: Remove KB as a suffix from ALL Melee weapons and make KB be something you can upgrade. This works perfectly because the targeted Class/Merc uses 2-Handed weapons, which natively get 10 upgrades instead of 5. Sacrificing 1 of those for the stat by choice would be the perfect solution to this issue.

Slaminaiodsnips:
Hit blinds target also sucks due to the removing player provided curses with the "blind" effect.

Please remove both Knockback and Blind on target from ALL weapons.  Allow players to pick gear or an Item Upgrade as a choice to have such effects.  Neither effect should be put onto a weapon directly.

It's crazy when a BIS merc weapon is removing curses applying blind like on the runeword weapon PRIDE.

FWCrash:
> Hit blinds target also sucks due to the removing player provided curses with the "blind" effect.
> 
> Please remove both Knockback and Blind on target from ALL weapons. Allow players to pick gear or an Item Upgrade as a choice to have such effects. Neither effect should be put onto a weapon directly.
> 
> It's crazy when a BIS merc weapon is removing curses applying blind like on the runeword weapon PRIDE.

Absolutely true aswell. Altho i'd argue there are some weapons that are ok to spawn with this affix, but very specific and prolly mega time consumeable to determine those. Would be best to just remove them period and let players use a recipe to up their 2H weapons to obtain either

Delegus:
> Hit blinds target also sucks due to the removing player provided curses with the "blind" effect.
> 
> Please remove both Knockback and Blind on target from ALL weapons. Allow players to pick gear or an Item Upgrade as a choice to have such effects. Neither effect should be put onto a weapon directly.
> 
> It's crazy when a BIS merc weapon is removing curses applying blind like on the runeword weapon PRIDE.

Doesn't sound like a BIS weapon then.

> > Hit blinds target also sucks due to the removing player provided curses with the "blind" effect.
[...]

PR #930, opened 2025-08-02
the big ole re-itemization pr

The-Honorable:
# Goals
> Re-itemize
> Adjust specific stats down like crushing blow/deadly so they are harder to achieve high levels of. Re-tune items based on their level, cost to make(runewords). Tune down some more op items, and rework uniques to be more worthy an option to use at all stages of the game. 

## Armor base changes
- Dexterity requirement added to circlets(40), coronets(50), tiaras(60), and diadems(80)

## Weapon base changes
- Sockets added to throwing weapons
  - <ins>1 Socket: </ins>
    - Throwing Knife
    - Throwing Axe
    - Balanced Knife 
  - <ins>2 Sockets:</ins>
    - Balanced Axe
    - Javelin
    - Pilum
    - Short Spear
    - Glaive
    - Throwing Spear
    - Battle Dart
    - Francisca
    - War Dart
    - Hurlbat
    - War Javelin
    - Great Pilum
    - Simbilan
    - Spiculum
    - Harpoon
    - Flying Knife
    - Flying Axe
    - Winged Axe
    - Maiden Javelin
    - Ceremonial Javelin
  - <ins>3 Sockets:</ins>
    - Winged Axe
    - Hyperion Javelin
    - Stygian Pilum
    - Balrog Spear
    - Ghost Glaive
    - Winged Harpoon
    - Matriarchal Javelin 

## Rune Changes
- Ber Rune: Crushing Blow changed from 20 to 10

## New Itemstats
- extra_ravens: Increases maximum raven petmax. Added to several uniques.
- extra_spirit_wolves: increases maximum spirit wolf petmax. Added to several uniques.
- extra_dire_wolves: increases maximum dire wolf petmax. Added to several uniques.

## Monstat changes

### New Monsters added
- Green Widow added
  - High poison damage to attacks
  - Added to Arachnid's Bite unique as reanimate
  - Does not naturally spawn in the wild

- Voltshade added
  - A willowisp that casts its own chain lightning
  - Added to Wisp Projector unique ring as reanimate
  - Does not naturally spawn in the wild

## Skill changes
- Raven petmax adjusted to be based on level instead of base level allowing for CTC items and oskill Raven to be useful
- Grizzly stun duration reduced from 20 frames +5 per level to 12 frames at all levels
- Fire golem now inherits Fire sunder effect and the players % fire skill damage and - enemy fire resist stat

### Chance to cast specific skills
[...]

Issue #25, opened 2024-05-13
Add Corpse Explosion Effect


CollinHerber:
Right now it is extremely difficult to tell where the corpse exploded. Adding some sort of effect to visualize this happening would be nice.

PR #986, opened 2025-11-11
Random Nonsensical Tweaks

Delegus:
Resolves: #681 #25 

### Missiles
- Fix Winters Pulse reference

### Paladin
- Conviction given linear defense % debuff scaling instead of 0-100 in DM56 style.
  - Base 15% +1% per level and caps at 50%
- Holy Freeze
  - Added formula to movement speed and to other_animrate

### Necromancer
- After investigation corpse explosion already has extra effects built into it so for visual clarity I added the teeth explode to the cast, just a small white flash to show where the corpse exploded.
- Fixed tooltip for Skeleton Warrior displaying incorrect damage during skill allocation.
- Fixed tooltip for Skeleton Mage displaying incorrect damage during skill allocation.
  - Change to use descmiss and associated calcs to simplify life for everyone
- Fixed revives missing fire/cold/light sunder
- Changed revive tooltip to use clc1 for hp and ast1 for damage display is unchanged but life easier goals.
- Reorganized passivecalcs for golems
- Fixed golem tooltips from displaying incorrect all the things during skill allocation.
- Reorganized tooltips for golems, from top to bottom they now display "special stuff" "life" "damage" "attack rating" "defence" and on all but iron "golem count"
  - Had to combine Defense and Attack Rating into 1 string and pass it 2 calcs and do the same for the blood aura in order to get the tooltip to show properly with unallocated skill points.

Confuse Curse -> Corpse Field
- Spawns corpses
- 3 base, +1 per 5 lvl
- 5s CD base, -1s per 12 lvl
- Spawns a 1 HP mob with no treasure class and cannot be revived and is useless for CE.
- Replaced Confuse on items with Attract

Fixed Akara's Point -> Akara's Wrath

Big skill name pass through uniqueitems, setitems, and sets I had planned before someone started PR #930 
- Removed knockback from multiple set items in case we don't get to reworking them.

Added hirelinginventorypanelhd.json to pc/handheld/ because it borks sometimes on UMPCs


The-Honorable (approved):
lgtm

PR #994, opened 2025-11-24
Revise new Korean strings

d3m3vilurr:
additional note:
- all languages of `HpSynergy` are broken; it's `pdrSynergy` strings.
- strings of `Chillwind` is conflict `Frostwind`.

and I didn't check this patch set in the game. I just checked the updated strings (so, it can have a bug)


Delegus (approved):
Merging, noted details for others, Ben is going to fix em.

PR #1046, opened 2026-02-22
302 Bug Fixing Dele Style

Delegus:
- Fix Sanctuary causing hit recovery
- Fix Sanctuary pulling enemies towards you
- Change waypointspanel to use a definition instead of a reference due to clients in other languages failing to be the proper color on occasion
- Fix Summon Dire Wolf max pet count again.
- Fix Valkyrie Prerequisite to be Slow Missiles instead of Inner Sight
- Fix Runestar unique magic pierce
- Fix row layout of monstats.txt for demon pets
- Fix Chasm Break, Carnage, Tectonic Slam having no leech as this was not intended, it appears a change to missiles broke/fixed something else.
- Fix A3 merc not using buff spells properly

### Demonic Mastery
- Pet cap is 4
- Begins at 1 and gains +1 per 8 base points in any combination of Summon Goatman, Summon Defiler, Summon Tainted, and Demonic Mastery

### Blood Oath
- Damage transfer capped to 50%

### Summon Goatman
- Adjusted damage scaling so they benefit from tiered scaling via skill level curve instead of flat linear lvl*2 so they aren't helpless without crushing blow

### Summon Tainted
- Adjusted fire penetration to 5% + 1% per Demonic Mastery bonus level

# UI
- Update advanced tabs for MnK
image
image
image
No cube:
image
- Update advanced tabs for Controller:
image
image
image
No Cube:
image


The-Honorable (approved):
honk honk

PR #1056, opened 2026-02-26
Fix: adjust Hephasto (H) rune drop

syxxness:
Corrected Hell Hephasto to correctly drop the intended runes.


syxxness:
I'm going to close this an resubmit cleanly.

PR #1085, opened 2026-03-04
Cow WP (Act 4 Submenu)

GildyBoye:
- Adds Secret Cow level WP under A4 Submenu

image

image


yinyin333333:
> it works well. I just found one problem that when act4 is tz, moo moo farm wp is shown normal zone.

That’s because it’s missing from the Terror Zone waypoint list.
If you add "waypoint_level_id": 39 to "level_id": 39 in desecratedzones.json, the added waypoint will also be highlighted in purple during Terror Zones.

PR #1086, opened 2026-03-04
Various Bug Fixes

Delegus:
# Herald Tiers

Swap heraldtier to correct row in itemstatcost.txt by swapping places with stacked_terror_key so it updates correctly

> [!NOTE]
>Appears to be name only as they do seem to tier up properly in stats anyways but this may cause problems in the future and I would rather just fix it.

> [!WARNING]
> - Keys must be removed from keychain and the keychain discarded from all characters and personal and shared stash tabs before updating the mod to the new version.
> - Individual keys in any location or stash tab are completely fine, the "Keychain" item itself must not be anywhere in the save file or the character/stash will not load.

# UI
- Disabled grabbers and storage items from the Chronicle though it's bugged at the moment, when it does work, those are just redundant.
- Fixed cold sunder text appearing too low on items.
- Add missing sound effect for metalHitsound in translationsettings.json

### Battle Orders
- Scaling formula corrected

### Unstoppable Force
- Missile speed increased slightly

### Chasm Break
- Area of effect visual reduced to represent damaging area better

### Madawcs Fury renamed to Korlics Fury
- Cosmetic change to match the Uber Ancient elements

### Vengeance
- Skill reworked to operate differently behind the scenes.
- The AoE no longer procs CTC
- Finally remembered to add scaling to the base skill and remove some of that from the synergies.
  - 20 + 5 lvl, Holy Element Auras to 4, Conviction to 8

### Death Sentry
- Added magic mastery and pierce inheritance so it works properly.
- Swapped the sound on hit

### Echoing Strike
- Remove flat damage scaling entirely... really SnowCo?
- Cap Projectiles to 5
- Fix formulas to not be overengineered for their purpose.

### Summon Tainted
- Added Fire tag

### Death Sigil
- Added Fire tag

# Cube Recipes
- Shards convert directly into the next shard like grabbers
- Statue Conversion Added

```
Act Statue + Different Act Shard = Shard Act Statue
Act 1 Materials - Eastern Worldstone Shard & Talic's Anguish Statue
Act 2 Materials - Western Worldstone Shard & Korlic's Pain Statue
Act 3 Materials - Southern Worldstone Shard & Madawc's Ire Statue
Act 4 Materials - Deep Worldstone Shard & Bul-Kathos' Nightmare Statue
Act 5 Materials - Northern Worldstone Shard & Worusk's End Statue

eg: Talic's Anguish Statue (A1) + Northern Worldstone Shard (A5) = Woruk's End Statue (A5)
```

# Treasureclass
- Fixed odd drop rate decrease for orbs and jewelry going from N > NM > H
- Updated statue drop % from act bosses to 6% from 3% to match key bosses.

# Monsters

- Uber Ancients resistance penalty down to 150% from 200%
- Reduced Korlic Blizzard damage by 20%
- Reduced Madawcs Thunder Storm damage by 20%
- Updated hireling monsters to restore some vanilla values when changes appear to be random experimentation at some point in the past.
- Reduced density in side areas affected by 2.2.0 boss testing merge
> [!NOTE] 
> - Yeah this is popular but it was unintended and the mobs can't even move or barely defend, they are just loot pinatas, it was reduce the density or make them much stronger and flying.
> - In exchange I have not reduced it entirely back to 2.1.2 but left is about 25% higher instead of double, in addition I have increased multiple other main and side areas.

### AI fixes by CantripN

- Necromancer Skeleton Warriors and Mages adjusted to be more responsive to nearby enemies

# Terror Zones
- Heralds now tier up properly in name and loot and not just difficulty
- Herald spawn chance increased to 5% from 2%
- Harold Holy Freeze damage has been fixed to not be 3x what I thought it was.
- Act 1 removed Stony Field and Tristram
- Act 4 gained Stony Field and Tristram

> [!NOTE] 
> Not Patch Notes
> - Removed flags on splash charm placed in error in the past
> - Fix misspelling on El'Druin of FotH (works in game but docgen hates it so it's being fixed. 🤣 )

### DIFF
--- a/data/global/excel/base/runes.txt
+++ b/data/global/excel/base/runes.txt
@@ -1,7 +1,7 @@
 Name	*Rune Name	complete	disallowCraftingInLadder	disallowCraftingInNonLadder	firstLadderSeason	lastLadderSeason	*Patch Release	itype1	itype2	itype3	itype4	itype5	itype6	etype1	etype2	etype3	*RunesUsed	Rune1	Rune2	Rune3	Rune4	Rune5	Rune6	T1Code1	T1Param1	T1Min1	T1Max1	T1Code2	T1Param2	T1Min2	T1Max2	T1Code3	T1Param3	T1Min3	T1Max3	T1Code4	T1Param4	T1Min4	T1Max4	T1Code5	T1Param5	T1Min5	T1Max5	T1Code6	T1Param6	T1Min6	T1Max6	T1Code7	T1Param7	T1Min7	T1Max7	*eol
 Friendship	Friendship	1						club	hamm	mace							HelAmn	r15	r11					levelreq		5	5	addxp		3	3	rip		1	1	dmg%		133	133	dmg-norm		25	50	crush		10	10	deadly		15	15	0
 Ancient's Pledge	Ancient's Pledge	1						shld									RalOrtTal	r08	r09	r07				res-cold		30	30	res-all		13	13	ac%		50	50	dmg-to-mana		10	10													0
-Armageddon	Armageddon	1						swor									RalOrtThulChamBerHel	r08	r09	r10	r32	r30	r15	levelreq		13	13	dmg%		275	300	dmg-elem	200	50	300	hit-skill	Blizzard	15	20	hit-skill	Armageddon	15	20	swing2		30	30	res-all		35	35	0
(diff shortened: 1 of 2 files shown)


The-Honorable (approved):
yeetus goosus

PR #1092, opened 2026-03-08
Barb: allow dual-wield for staves, polearms, mauls, and axes

yinyin333333:
Extends Barbarian dual-wield to additional 2H weapon types (staves, polearms, mauls, axes) for consistency with dual 2H swords

I've set the one-handed damage to approximately 50%. Since it also allows users to pair a Maul with a shield even without dual-wielding, it may not be a massive change; however, I'm submitting this PR in hopes that it helps expand options for Barbarian players in a positive way.

And modified the state machine JSON because the frontend does not recognize when a weapon is held in each hand. By changing the neutral animation of stf to that of 1hs, the character selection screen now correctly displays the weapons being dual-wielded.

test video
https://www.youtube.com/watch?v=RYDv7ga4pBw


Delegus:
Closing this can revisit in the future if needed.

PR #1095, opened 2026-03-09
Fix Cow WP level/lvlprest

yinyin333333:
Adjust levels.txt and lvlprest.txt for Cow WP behavior

In 3.0.3, the Cow WP had scan disabled, which caused it to warp players to a random location instead of the actual waypoint. However, simply enabling scan is also problematic because players get surrounded by cows as soon as they land. Therefore, I've enabled scan but adjusted the settings to ensure players remain safe when using the waypoint.

gif1

this 3.0.3 (adjust enable fixed waypoint area)

gif2

and apply this PR


yinyin333333:
In most cases, a warpdist of 1024 is sufficient, but I've increased it to 2025 to ensure better stability. I also slightly shifted the Cow King and Cow Queen to the left within the ds1 file to provide additional stability.

Delegus (approved):
Seems fine after a bunch of in and out between NM and Hell

PR #1096, opened 2026-03-09
Revert Mod Menu Localization

Delegus:
Localization changes cause crashing on ruRU language and are thus being reverted. I tried multiple different methods but short of blanking out the entire ui.json file or removing the localization calls I cannot find a solution. Investigation shows language character limits inherit to D2R, what a big suprise.


d3m3vilurr:
that's too bad.... :'(

The-Honorable (approved):

PR #1101, opened 2026-03-10
Apply rune name translations on mod info tab

d3m3vilurr:
we cannot use `rXX` key. these have `color` tag.
another key `rXXL` only have plain text

----

these rune names is defined in `item-runes.json` similarly gem names & gem strings of `item-names.json`

I also know #1096 PR. IDK this strings are safe or not in other regions.. only tested with korean.

D2R_TNfSsiDmOT

### DIFF
--- a/data/global/ui/layouts/modinfo_runetablehd.json
+++ b/data/global/ui/layouts/modinfo_runetablehd.json
@@ -130,7 +130,7 @@
                                     {
                                         "type": "TextBoxWidget", "name": "Rune Name",
                                         "fields": {
-                                            "text": "El (1)",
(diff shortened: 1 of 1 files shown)


Delegus:
We're removing menu support from main mod, that being said the best method prior to that being a thing would be to make a new key.

d3m3vilurr:
del explained to me about [detail issue](https://us.forums.blizzard.com/en/d2r/t/modding-crashes-for-russian-language/168281)

Delegus:
Actually just closing this for now so it isn't merged on accident.

PR #1102, opened 2026-03-10
Delete Mod Item Chronicle Flags

Delegus:
- Add 'disableChronicle' to all mod-added rows to prevent crashing on item identification.
- > After testing a while I do know vanilla rows appear to work fine so left them enabled.


The-Honorable (approved):
bionicles > chronicles

PR #1105, opened 2026-03-11
update zhTW string

kdsneo7:
Eng:
1. Fix some wrong words in zhTW.
2. add English Sets name after zhTW Sets name, for easier serch new Sets on website.
3. add Bases tire and type words.

中文:
1. 修正繁體中文字串
2. 於中文套裝名稱後面補充英文版套裝，方便於D2R Reimagined網站查詢相關資訊
3. 基礎裝備名稱後方補充物品階級以及重量類型


Delegus:
UI.json is out of date, this had to be reverted to not support localization. Your item-names.json breaks formatting and causing whole file to flag as diff; May be changing from CLRF to LF line breaks, the first 2 I noticed using GitHub desktop as it won't load here.

kdsneo7:
that okey, I will download file from here and reupload

PR #1106, opened 2026-03-15
Fix Warlock Mana Regen

Delegus:
Title


Bakelt (approved):
Should have brought life tap

PR #1109, opened 2026-03-17
Necro Poison Dagger -> Strike, Reverting Life & Death to Former Status

GildyBoye:
Necro Poison Dagger renamed to Poison Strike, opened up to other melee weapons.
- Necro and Warlock both roll on dagger skills etc (not awful but kinda ass when it's the only wep type you can use with the skill)
- Opens up the skill to be used with various builds, items and sets, especially while leveling without focusing on 1 specific item type.
- 2h builds such as Scythes etc will be usable.
- 1h/2h runewords/uniques outside of daggers can also be used for build options and variety, examples such as Harvest Moon and Grief.

--------------------------------------------------------------------------------------------------------------------------------------

image

Life and Death returned to former status so its usable.

- ASK 1-1
- Life Drain and Life % Removed
- MF/GF reduced to 25-50


yinyin333333:
I tried out the skill with a Thresher, and it fits really well.

Delegus (approved):
Welcome back to your boring generic pointless lane L&D

PR #1110, opened 2026-03-18
Fix G Orb Arrow Spawner

Delegus:
- update collision so it properly spawns when hitting walls and enemies near walls


Bakelt (approved):
LGTM

PR #1112, opened 2026-03-19
Bug Fixes Druid and Warlock Tweaks

Delegus:
### Glacial Orb Arrow
- Fix overly aggressive NHD preventing all 3 nova from hitting 1 target.

### Summon Spirit Wolf
- Twister skill tooltip updated to reflect correct damage.
- Chill limit of the pet actions set to above where actions stop entirely.

### Summon Fenris
- Firestorm skill tooltip updated to reflect correct damage.
- Chill limit of the pet actions set to above where actions stop entirely.

### Summon Grizzly
- Pet damage growth has been placed on the correct damage type. No longer flat instead of %
- Curve increased slightly
- Shockwave skill tooltip updated to reflect correct damage.
- Chill limit of the pet actions set to above where actions stop entirely.

### Solar Vine
- Removed bugged and unworking monster property

### Fist of the Heavens
- Now scales holy bolts with magic mastery

### Ring of Fire
- Level 23+ damage curve increased

### Flame Wave
- Level 23+ damage curve increased

### Ehanced Entropy
- Gains 1% magic skill damage per base level.
- Grants Miasama Bolt an extra projectile every 5 base levels up to 5 total.
- All skills relying on this passive for AoE and duration have been made into bonus growth instead of requiring base points.
  - Miasma chains and bolt cloud and duration and abyss size now correctly gain the bonuses the tooltip of this skill shows at the time it shows them.

### Abyss
- Base radius changed from 7-9 to a flat 8.
- Cooldown increased slightly up to 1 from 0.6 second.
- Level 23+ damage curve increased

### Summon Tainted
- Synergy with Blood Boil now works properly, tooltip shows the correct damage, and synergy is 15%

### Sigil Lethargy, Rancor, Death
- Radius increased slightly 6/8/10 at level breakpoints from 4/6/8 to better match visual.
> Making the missile itself smaller/larger breaks the visual so yay.

# Items

## Vendors
- Cleaned up random large number of duplicate potions
- Removed erroneous items from staff mod vendors

## Uniques
- Fixed randomclassskill on Spirit Ward and Gambler's Fallacy to include Warlock
- Fixed vanilla renewed sunder to fixed penalties to reflect vanilla
- Leader of the pack removed range from requirement penalty, is now always 100%
- Heaven Facet now only gives +1 all skills, removed flat HP and Mana
>This jewel really is annoying, if not for summoners I would just replace it.

### 5 item unique/set recipe
- Now accepts a combination of any 5 weapon/armor pieces + ID tome instead of requiring matching bases.


data/local/lng/strings/skills.json
```diff
@@ -39663,19 +39663,19 @@
...
-        "ptBR": "Tempestade Ígnea Dano ígneo: %d-%d",
-        "ruRU": "Огненная буря Урон от огня: %d–%d",
-        "zhCN": "火焰风暴 火焰伤害：%d - %d"
+        "enUS": "Firestorm Avg. Fire DPS: %d-%d",
+        "zhTW": "火焰風暴 平均火焰每秒傷害：%d-%d",
+        "deDE": "Feuersturm Ø Feuerschaden pro Sek.: %d-%d",
+        "esES": "Tormenta de fuego DPS medio de fuego: %d-%d",
+        "frFR": "Tempête de feu DPS de feu moyen : %d-%d",
+        "itIT": "Tempesta di Fuoco DPS medio da fuoco: %d-%d",
+        "koKR": "화염폭풍 평균 화염 초당 피해: %d-%d",
```
d3m3vilurr:
```suggestion
        "koKR": "화염폭풍 평균 DPS: %d-%d",
```
:)

Bakelt (commented):
Only bit that looks odd to me is Dsum_Firestorm going from a hitshift of 2 to 8 is going from 4/256ths to 256/256ths while SrcDmg is only going down from 12/128ths to 8/128ths.

Bakelt (approved):
LGTM

Can't test warlock stuff but don't care about warlocks so yolo

PR #1118, opened 2026-03-23
affix fix

The-Honorable:
- fix life affix that was causing sunders to roll 99
- enable the random lower tier affixes that werent able to roll on crafted charms
  - seemed odd that the higher lvl req affixes of the same prop could roll but not the lower tiers even tho some of them have max level set


Delegus:
Because the pool is bloated enough and the crafted charm level has no variety but 60 or 80

Delegus:
Added the vita change to my PR since you is out of the PC for the night.

PR #1119, opened 2026-03-23
quick fixes/tweaks

Delegus:
Update Lethargy Tooltip
Update Blood Oath Tooltip
Fix Glad Bane
Revert L&D Change
Make Shards useable in NM after Baal Kill
Remove GC affix from crafted charms causing Sunder to hit 99 required level.
Update AI expectancy for new traps
NHD on Twister to 10
NHD on Tornado to 15
Disable some random affix pet CTC that can crash
Lower Firestorm light


Bakelt (approved):
LGTM

PR #1126, opened 2026-03-24
Return standard of heros to ques item from key

d3m3vilurr:
this patch will resolve unable unlock steam achievement issue. I don't remember this item is stackable or not
(sorry I don't have key enought to test)
at least, we can still use rotw stash.

image

image


Delegus:
Does it break anything on no dlc for putting it on the shared stash itself pages?

d3m3vilurr:
well... delegus shares to me another problem.
ques item cannot be sold; means cannot summon dclone...

this patch cannot merge until finding option of this problem 😢

CollinHerber:
Couldn't an item be made that isn't a quest item that functioned the same as SOH?

d3m3vilurr:
in discord ideas

1. cube SOH to SOJ (or make another ticket item)
2. make a new map for dclone then cube SOH to open a portal

and also Del shares info, that ubers drop the hardcorded items; SOH and CM2

I made pr for achievement, so for this purpose, SOH should set a `ques` type.
IMO, a client is hardcorded that obtaining quest item unlocks achievement.
that these reason we cannot use `similar item as SOH` option to keep current flow.

yinyin333333:
> in discord ideas
> 
> 1. cube SOH to SOJ (or make another ticket item)
> 2. make a new map for dclone then cube SOH to open a portal
> 
> and also Del shares info, that ubers drop the hardcorded items; SOH and CM2
> 
> I made pr for achievement, so for this purpose, SOH should set a `ques` type. IMO, a client is hardcorded that obtaining quest item unlocks achievement. that these reason we cannot use `similar item as SOH` option to keep current flow.

gif1

I'm not sure exactly what method you're looking for, but is this what you mean?
The cube recipe ingredients are just for testing, so I used ISCs (Identify Scrolls), and I've placed Diablo Clone in a different map.
The map itself is a temporary one where I flattened Tristram into a plain area last time to use for Reimagined for D2RLAN.

Delegus:
While cool, I prefer his vanilla spawning behavior. Opinion is just re-use or add a misc.txt item that SoH can cube into or be transmogrified into to be sold to spawn him and save the custom levels and portals for mod-added bosses in reference to retail/offline.

yinyin333333:
i agree too. While placing it in a different map certainly has its own meaning and appeal, many people are so used to the process of selling a misc item to a vendor and then going out to find Diablo Clone that it might cause quite a bit of confusion. (Of course, people will adapt to it over time, though.)

PR #1128, opened 2026-03-24
Allow Respec without losing Bind Demon

Cantrip-commits-crimes:
Sets Base Max to 1 so you can respec without losing Bind Demon. Cap is still always 1.

Previously even having an item with the skill made it die when doing a respec.


Delegus:
Considering this is a direct introduction of an exploit to re-spec and have a permanent bound demon with snapshotted stats that saves between games, this gets a hard pass from me.

Cantrip-commits-crimes:
Ah, apparently this doesn't work if you S&E. If you S&E without a single point in Bind Demon it vanishes.

Snapshotting levels and such already works in vanilla, this doesn't add to that. At most you can get a few extra points for a boss fight, then respec again to re-allocate at the end of a session.

syxxness:
Just make Hephasto always spawn with cursed affix :smiley:

The-Honorable:
pretty much agree with del. iron golem also has a similar thing and I think its fine.

ProdigyWTF:
Nope to any of this. Don't add exploits to a busted char.

PR #1131, opened 2026-03-25
Fix minor

### LINKED ISSUE #1129: Quiver/Bolt of Slaying Not Aligned
Quiver is 25% pierce
Bolt is 15% pierce

Bolt is advertised as 25% on wiki

### LINKED ISSUE #1130: A3 Merc Icicle Barrage Collision
Collision uses flamewave as a base which means it fails to go over gaps like rivers or arcane sanctuary abyss.

Delegus:
Resolves #1129 
Resolves #1130 

Fix Crystal Sword not showing ilvl
Fix pet display order
Fix Bolt Case of Slaying pierce
Fix Ice Wave collision to float and casting visuals to ice based
Fix the other A3 merc spells to do ice/fire/lightning respectively
Fix Shaman in normal Baal wave 1 fire immunity
Fix Sigil of Lethargy showing a positive value for enemy PDR
Fix Frost Nova showing enchant damage added to the missile
Fix Crescendo not having a method of obtainment
- Quiver of Slaying + Ist Rune + Ral Rune + Rare Jewel + Ruby

Update "Damage Taken Goes to Mana" to "en Added to"

Proposing sorc armor changes to reduce free high defense % in line with other 2.2 skill changes
- frozen def base and per level down to 15 and 3 from 30 and 5
- shiver def base and per level down to 20 and 4 from 45 and 6
- chilling def base and per level down to 25 and 5 from 60 and 9
- buffed based PDR % to 5, still caps at 25% so the nerfs have a bit less sting and doesn't just do 1% 🙄


Bakelt (approved):
LGTM

PR #1133, opened 2026-03-27
Allows warlock to whirl the wind

GildyBoye:
Allows warlock to properly spin

**BEFORE**
HeDontSpin

**AFTER**
HeSpin


Cantrip-commits-crimes:
I like the before better, tbh. Look more menacing and eldritch.

Delegus (approved):

PR #1134, opened 2026-03-28
Strafe to All/Lock Zeal Fix

Delegus:
Update Warlock Zeal to include the other 1H and 2H weapon types it was missing
Add strafe state machine to all 7 non amazon cllasses.


Bakelt (approved):
I tested nothing and you can't make me

PR #1138, opened 2026-03-30
fixing a bug or two

The-Honorable:
- scorchveil hit shift correctly swapped to 4 instead o 8
- disabled darkfear in chronicle as it does not exist in game


Bakelt (approved):
LGTM - hitshift matches base skill (Fire Wall)

PR #1142, opened 2026-04-01
Unborked trickster pr

Cantrip-commits-crimes:
Trickster Re-PR since borked on merge with other minion PR by Del.

Tricksters replace Defilers, art/sound/summon icon, all of it (still missing skill icon).

Why? Because Defilers are ugly and look useless, while Imps are adorable, do a cool backflip, and are just way more fun to use.


Delegus (approved):
Things work, stuff dies, no one objected. LGTM

PR #1144, opened 2026-04-01
Patch 3.1.2 parity

Delegus:
Just random housekeeping looks like.

monstats.txt may change how the ancients fight works out, the 3 is undocumented in that column but I didn't notice anything. 🤷


The-Honorable (approved):
katie parity

PR #1146, opened 2026-04-01
More lint updates

Delegus:
Add json key/id duplicate check across all present local/lng/strings json files
Checks for keys not in use beyond a user set threshold (find unused strings or key matching typos in .txt and layout .json)

File Name and entries removed because not in use or duplicated:
- item-modifiers.json
  - -45
- item-nameaffixes.json
  - -1
- item-names.json
  - -3
- levels.json
  - -9
- skills.json
  - -8


The-Honorable (approved):
shirly nothing can go wrong right?

PR #1149, opened 2026-04-03
Gambling Skill Gold Adjustments

Delegus:
- Adjust skill gold multiplier and add to prevent early game infinite gamble but still allow class items to be worth something late game.


The-Honorable (approved):
please the gheed

The-Honorable (approved):

PR #1150, opened 2026-04-04
Gheeds Greed

Delegus:
- update items with inherent magic level to a higher base cost
- remove the cost bonus to hidden charm passive I accidently left in on the other PR while testing
- set all instances of rep-quant to 25 to hopefully squash the no sell issue for good
  - after much investigation rep-quant property is actually bit limited to 0-63 but values above 25 don't do anything as it still ticks up every 4 seconds or so
  - every affix up to this point was set to 100 so it was rolling over and unless set to specific values in cost/add causing unsellable until S&E in many cases


The-Honorable (approved):
repreprep

PR #1153, opened 2026-04-04
A3 Cold Merc Freeze Fix

Cantrip-commits-crimes:
Glacial Spike will Freeze now


Delegus:
Update skilldesc.txt row 57 column AJ to `len` (auralengthcalc pointer) and verify in game on a sorc and an A3 merc the tooltip displays the correct duration for both.

image

Delegus (approved):
Feels alright in game. Tooltips line up properly.

PR #1157, opened 2026-04-05
feat: lint update to check json for missing languages

Delegus:
We learn from mistakes sometimes


The-Honorable (approved):
belly button lint

PR #1161, opened 2026-04-06
Remove terror zone darkening

yinyin333333:
Removes the darkening effect from terror zones.
also removed the "thump-thump" sound—kind of like the Netflix intro sound—that plays whenever you enter a Terror Zone area.


Delegus:
Act 1:
- Does not display correctly alone because Stony Field is missing. This file sets the threshold to 90% required instead of 100%
     [levelgroups.txt](https://github.com/user-attachments/files/26523940/levelgroups.txt)

- Alternatively move Stony Field and Tristram back to Act 1 in the desecratedzones.json since we have a cow WP they aren't really doing much in Act 4 anymore.

Run lint and fix the levels.txt vis0 error and the levels.json errors. All of that is in the rescue PR anyways. Can ignore the monster chain stuff, lolz.

Delegus (approved):

PR #1163, opened 2026-04-07
Bind Demon Exile

Cantrip-commits-crimes:
Bind Demon pushed outside the pet count for Warlock.

Design Intent: No FOMO from using Bind Demon on a summoner lock, vs fire/magic/eldritch having no opportunity cost for it.

Side Effect: Not able to Consume Bind Demon (good, no missclick), not able to Death Mark (good, doesn't get stuck inside Summon Tainted).

Changing Tricksters to be their own type disabled Consume for them as well, which is not acceptable since Magic Lock exists and wants to eat them, so this is the next best thing.

---

Summon Tainted: Synergy further nerfed to 8% from 10% to compensate.


Cantrip-commits-crimes:
No use looking into this until seeing the impact of the PTR changes on it.

Tricksters # of Links per level is worth raising imo given the incredible density, though.

PR #1166, opened 2026-04-08
amazon pet suck make gud wif aura

Cantrip-commits-crimes:
Valkyries always spawn with a level 3 aura from: Concentration (for phys attack), Fanat (for elemental attack), Sanctuary (for magic).

Replaces any armour they'd randomly get normally, so they never have more than this aura, just a level 3 aura on a chest with no other mods (unique for them, drop disabled - I think).


Delegus (changes requested):
A few issues in formatting:

1) Why is this based in misc.txt where it is not assigned the usual base armor values and thus provides nothing defensively, is there a limitation of some kind?
2) Why is the code 'valk' ? D2R and 4 letter codes generally cause a lot of mischief and random problems, even the data guides list it as 3. While it works in some applications, it is best practice to use 3 as every other base item and misc item in the parent game and the mod do.
3) Why are there 3 new armors when 1 will suffice with a property group, limitation of some kind?
4) The IDs are incorrect on the uniqueitems.
5) The property "aura" has been updated to use name instead of ID # in uniqueitems.txt for the simplicity of searching and reading skills. 
    >We still need to do the same in other files but they haven't been overhauled yet like uniques were in 2.2

yinyin333333:
I suspect the reason he based it on misc and didn't use property groups is that my previous example for monequip was based on a version without property groups using a ring from misc, and he likely just followed that as-is.

Delegus:
> I suspect the reason he based it on misc and didn't use property groups is that my previous example for monequip was based on a version without property groups using a ring from misc, and he likely just followed that as-is.

Ah so the bad influence reveals itself. 🤣

yinyin333333:
> Ah so the bad influence reveals itself. 🤣

Hahaha. May the grace of RotW be with us all. (Though I, of course, have not been blessed by it.)

Cantrip-commits-crimes:
> A few issues in formatting:
> 
> 1. Why is this based in misc.txt where it is not assigned the usual base armor values and thus provides nothing defensively, is there a limitation of some kind?
> 2. Why is the code 'valk' ? D2R and 4 letter codes generally cause a lot of mischief and random problems, even the data guides list it as 3. While it works in some applications, it is best practice to use 3 as every other base item and misc item in the parent game and the mod do.
> 3. Why are there 3 new armors when 1 will suffice with a property group, limitation of some kind?
> 4. The IDs are incorrect on the uniqueitems.
> 5. The property "aura" has been updated to use name instead of ID # in uniqueitems.txt for the simplicity of searching and reading skills.
>    > We still need to do the same in other files but they haven't been overhauled yet like uniques were in 2.2

1 - Can move to uniques, just copied the example given to me using a Ring.
2 - So, Val.
3 - Didn't know about prop groups, can do it as 1 then.
4 - Will look into it.
5 - Noted, will fix.

PR #1167, opened 2026-04-08
Paladin tooltip fixes

strangemolars:
Submiting only the paladin tooltip fixes I done. Wanna do them by themselves so they are not tied to other stuff I wanna fiddle with.

Changes made:
Holy Fire, Holy Freeze and Holy Shock: changed so the "damage added to your attack" part of the auras react to %elemental damage and show the values relative to the "first dip" of the %elemental damage increases.

Fixed Blessed aim passive effect showing as 25% before putting the first point, instead of just showing the 5% it actually gives.


strangemolars:
Freaking thing pulled some other changes I though I had already commited and deleted xD

PR #1168, opened 2026-04-08
Paladin tooltip fixes

strangemolars:
Submiting only the paladin tooltip fixes I done. Wanna do them by themselves so they are not tied to other stuff I wanna fiddle with.

Changes made:
Holy Fire, Holy Freeze and Holy Shock: changed so the "damage added to your attack" part of the auras react to %elemental damage and show the values relative to the "first dip" of the %elemental damage increases.

Fixed Blessed aim passive effect showing as 25% before putting the first point, instead of just showing the 5% it actually gives.


data/global/excel/base/skilldesc.txt
```diff
@@ -101,23 +101,23 @@ might	2	1	1	2	4	11	skillname98	skillld98	skillld98	skillan98																	36
...
 charge	1	3	3	1	22		skillname107	skillld107	skillld107	skillan107	7	ln34+(skill('Vigor'.blvl)+skill('Might'.blvl))*par8											2				74	StrSkillDamagePercent		ln34+(skill('Vigor'.blvl)+skill('Might'.blvl))*par8		74	StrSkill22		toht							75	StrSkillAddedFlatPhys		pnma	pxma											36	StrSkill3		usmc	256																					40	Sksyn	skillname107	2		76	Damplev	skillname115	par8		76	Damplev	skillname98	par8																								0
-blessed aim	2	3	1	2	24	8	skillname108	skillld108	skillld108	skillan108																	36	StrSkillRadiusSingular	StrSkill18	ln12*2	3	74	StrSkillAttackRatingPassive		(skill('Blessed Aim'.blvl) < 1) ? par8 : skill('Blessed Aim'.blvl) * par8		74	StrSkillAttackRatingActive		ln34																																																																															0
```
Delegus, line 110: Why did this change? Edited: LAWL

Ah I read contributor chat. This is happening because the formula is missing a set of ()'s.
`(skill('Blessed Aim'.blvl) < 1) ? par8 : skill('Blessed Aim'.blvl) * par8` essentially says `(blvl < 1 ? par8 : blvl) * par8` because only the first operation/parameter fed to : without parenthesis is considered within the statement. Alternatives are `(skill('Blessed Aim'.blvl) < 1) ? par8 : (skill('Blessed Aim'.blvl) * par8)` or changing skilldesc to `pst1` and skills.txt to `min(blvl*par8,par8)`

A very quick look at skills.txt because I am not chasing this 🐰 🕳️ shows it may be an issue for every aura whose passive is different than the allied bonus.

strangemolars: Just adding as a reminder - so confusion dont rise again when merging:

The formula (skill('Blessed Aim'.blvl) < 1) ? **1** : skill('Blessed Aim'.blvl) * par8 is a fix already. 

The number in bold just means it will show the par8 value as if it were lvl 1 (5%, basically showing as preview of what you will get putting the point, same as with the resist fire/cold/lightning passive max res bonus that use the same formula).

Delegus (approved):

PR #1171, opened 2026-04-10
Amazon: Changed Inner Sight and Slow Missile to use IAS

syxxness:
## Amazon: Inner Sight and Slow Missile to use IAS rather than FCR.

- "Inner Sight" and "Slow Missile" now scale speed of of IAS rather than FCR.  
- The extra stat requirements of using "Titan's/Elemental Covenant" makes extra FCR investment annoying and difficult when combined with the Amazon's poor FCR scaling.
- This change makes using "Inner Sight" and "Slow Missile" much easier to use in higher level area where monsters frequently spawn with speed modifiers and very high density.


Delegus:
This is attack animations controlled by cast speed in this setup. First column is animation used, second column is frame data used. SC is cast, A1 A2 is IAS. For now at least make it A1/A1.

syxxness:
Changed to A1/A1.  Tested again and seems to be working fine.

Delegus (approved):
Works fine, if a bit dopey looking.

PR #1173, opened 2026-04-11
fix: assassin crashing

Delegus:
shadow warrior attempts to cast shadow master when they do not share a pettype so changing master back to the base game as well as their ai types back to the base game


yinyin333333:
I followed the reproduction steps exactly and confirmed that the crash occurs (summoning Shadow Warrior while Shadow Master is assigned to a skill slot).

After replacing the files with the ones from the current commit, I’ve been hunting continuously to see if it triggers again, but so far, no crashes have occurred.

The-Honorable (approved):

PR #1178, opened 2026-04-13
Hello Lab, Bye Lab

yinyin333333:
PRing this before I forget.  This is not the final version, but a very rough and messy draft.

Adding Jail, Worldstone, and Nihl as mentioned by Ben to the existing Lab.

1~4 act3 dungeon
5 - lab5 ?
6~9 act1 jail
10 - lab10 ?
11~14 act5 Nihl
15 - Skip now
16~19 act1 cata
20 - Skip now
21~24 world stone (24 must be crash)
25 none


yinyin333333:
In the existing Lab5 ds1 file, the exit area wasn't properly blocked off, allowing players to teleport past it. I've added some tiles to fix that.
For the Lab10 file, I simply placed the tiles for the next exit.

few issues during a quick test:
I’m not sure if it’s because D2R dislikes having four consecutive mazes or if it’s just because I rushed the job, but while three consecutive levels seem fine, four mazes feel somewhat unstable.

I think it might be better to use Cata. I only tested it briefly, but Cata and Jail have such similar vibes that it felt tedious even in that short time. I feel like the Ice environment from Act 5 would look much more refreshing.

Also, I'm not sure if we strictly need to stick to the 4 levels + 1 boss room structure.

Whether we use an existing boss room or create a new one isn't the main issue; the difference between 4 levels and 3 levels is just one floor, yet it still feels tedious.

Delegus:
Should convert to draft so it is not merged by accident if it isn't ready. These things happen.

The-Honorable:
> In the existing Lab5 ds1 file, the exit area wasn't properly blocked off, allowing players to teleport past it. I've added some tiles to fix that. For the Lab10 file, I simply placed the tiles for the next exit.
> 
> few issues during a quick test: I’m not sure if it’s because D2R dislikes having four consecutive mazes or if it’s just because I rushed the job, but while three consecutive levels seem fine, four mazes feel somewhat unstable.
> 
> I think it might be better to use Cata. I only tested it briefly, but Cata and Jail have such similar vibes that it felt tedious even in that short time. I feel like the Ice environment from Act 5 would look much more refreshing.
> 
> Also, I'm not sure if we strictly need to stick to the 4 levels + 1 boss room structure.
> 
> Whether we use an existing boss room or create a new one isn't the main issue; the difference between 4 levels and 3 levels is just one floor, yet it still feels tedious.

can definitely edit the amount of levels inbetween bosses. nothing has to be the way initially described. thanks for the initial setup.

players are still functionally able to skip the boss though since the door is unlocked/accessible? would be a neat idea to somehow prevent that in the future.

yinyin333333:
> Should convert to draft so it is not merged by accident if it isn't ready. These things happen.

I'm not sure if I set it up correctly. I think I might have clicked something like "draft."

The-Honorable:
> > Should convert to draft so it is not merged by accident if it isn't ready. These things happen.
> 
> I'm not sure if I set it up correctly. I think I might have clicked something like "draft."

looks in draft mode now

yinyin333333:
> can definitely edit the amount of levels inbetween bosses. nothing has to be the way initially described. thanks for the initial setup.
> 
> players are still functionally able to skip the boss though since the door is unlocked/accessible? would be a neat idea to somehow prevent that in the future.

Yes, it's currently possible to skip them.
I wasn't sure how you wanted to structure it, so I just set it up that way for now.
The same goes for Lab10. It was originally a one-way map with no exit, so I just opened up an exit for the time being. We could make it so you transition via a portal, or... well, I'm sure there are several other ways.

Regarding the tile fixes in Lab5, that's separate from the boss issue; I blocked it off because you could literally go out of bounds beyond the walls. It's usually fine, but if the areas around entrances and exits aren't properly sealed, it can sometimes cause crashes. Visually, there’s no difference.

The-Honorable:
> > can definitely edit the amount of levels inbetween bosses. nothing has to be the way initially described. thanks for the initial setup.
> > players are still functionally able to skip the boss though since the door is unlocked/accessible? would be a neat idea to somehow prevent that in the future.
> 
> Yes, it's currently possible to skip them. I wasn't sure how you wanted to structure it, so I just set it up that way for now. The same goes for Lab10. It was originally a one-way map with no exit, so I just opened up an exit for the time being. We could make it so you transition via a portal, or... well, I'm sure there are several other ways.
> 
> Regarding the tile fixes in Lab5, that's separate from the boss issue; I blocked it off because you could literally go out of bounds beyond the walls. It's usually fine, but if the areas around entrances and exits aren't properly sealed, it can sometimes cause crashes. Visually, there’s no difference.

if there is a way to force the boss to drop a portal which leads to the next set of levels i think that'd be the best way. but it is something we can visit later. 

as long as the lab level functions that is a HUGE step forward since the initial lab was only setup for 10 levels and not the full 25 ish.

yinyin333333:
> if there is a way to force the boss to drop a portal which leads to the next set of levels i think that'd be the best way. but it is something we can visit later.
> 
> as long as the lab level functions that is a HUGE step forward since the initial lab was only setup for 10 levels and not the full 25 ish.

Since mazes literally depend on D2R generating them automatically, I actually didn't do much of anything. I just added the levels.
However, levels don't like sharing the same offsets. This can sometimes cause overlapping maps or lead to crashes.
using the base tilesets makes things much simpler. (Though it does make it feel a bit more... familiar, I guess.)

I'm actually starting to think this:

Does every level strictly need to be connected?
For example, we could create 4 or 5 portals within a space like Lab5. (Or we could place entrances in various underutilized areas of Act 5 to encourage players to explore different maps.)
Also, how about making each entrance lead to a different "Act-themed" environment?
That way, players could experience 3 or 4 different vibes, collect certain materials from each, and then transmute them to enter the final end-game content.

The-Honorable:
> > if there is a way to force the boss to drop a portal which leads to the next set of levels i think that'd be the best way. but it is something we can visit later.
> > as long as the lab level functions that is a HUGE step forward since the initial lab was only setup for 10 levels and not the full 25 ish.
> 
> Since mazes literally depend on D2R generating them automatically, I actually didn't do much of anything. I just added the levels. However, levels don't like sharing the same offsets. This can sometimes cause overlapping maps or lead to crashes. using the base tilesets makes things much simpler. (Though it does make it feel a bit more... familiar, I guess.)
> 
> I'm actually starting to think this:
> 
> Does every level strictly need to be connected? For example, we could create 4 or 5 portals within a space like Lab5. (Or we could place entrances in various underutilized areas of Act 5 to encourage players to explore different maps.) Also, how about making each entrance lead to a different "Act-themed" environment? That way, players could experience 3 or 4 different vibes, collect certain materials from each, and then transmute them to enter the final end-game content.

that also sounds like a good idea. Panda's initial goal was to re-work side areas to have more and harder enemies and maybe a boss in them so it is worth going into them as well. can certainly add a lot of different things in different areas that players can explore as well. players always like more content.

The-Honorable:
can confirm entering 24 crashes.

yinyin333333:
> for lab 16-19 i think swapping it to the jail layout would be nice instead of re-using cata. unsure if we can use the act5 frozen cave areas or not for maze too.

We can use the ice areas. Using them isn't an issue in itself, except for the fact that they branch out into entrances, exits, and additional exits leading to other blocked spaces (like the Frozen River).
While we could block those exits, it would likely affect other mazes and make things quite complicated.
And since levels 1 through 4 are already using the jail layout, is there a particular reason to use it again for levels 16 through 19?

> i do think to reduce the json amount we could easily just condense the level name strings to all just "Forsaken Labyrinth" instead of each level getting its own string.

Yes. However, I think it's better to leave the strings as they are while we're testing. That way, it'll be easier to identify exactly which level is causing the issue.

> for lab boss level 15/20/25 i think we could just re-used lab 5 and remove the door vis for lab 25 or make 25 a different zone i am not leaning any particular way for that. 

Yes, that sounds doable. The problem is that I haven't tried it yet, but I've heard that CTC Death can be a bit unstable. I believe there are cases where people spawn another monster when a boss dies and ensure that monster dies 100% of the time to mitigate this issue, but again, I can't say for sure since I haven't tested it. I'm also not certain if the portal missile will function correctly across all levels. If the portal to the next area fails to open even once when a boss dies, that would be a serious problem, so I think we need to give this some careful thought.

yinyin333333:
So, including the ones we haven't tested yet, there seem to be a total of three ways to transition from the boss stage to the next round.

1. Both entrances and exits exist.
If we go this route, players will be able to skip that level one way or another.
A small map like Lab 5, in particular, could essentially be skipped immediately.
On the other hand, the development difficulty is the lowest. Of course, some map tiles in certain acts don't play well with each other when connected, but in most cases, they work just fine.

2. portal is generated when the boss dies.
We eliminate the exit to the next floor and make it so that the only way to proceed is through the portal that appears when the boss dies. This would be the cleanest approach. However, as mentioned above, there seem to be some issues regarding Death CTC.

3. boss drops a Misc item. That Misc item then serves as a cubing recipe.
Yes yes, I'm not a fan of cubing recipes being used this way either.
However, this would be the most stable method.
It allows us to physically separate each level (5 and 6, 10 and 11, 15 and 16, 20 and 21) while maintaining a 100% success rate.
The downside is that you have to go back to town after clearing the boss floor to open the portal.
Yet, this drawback could actually be turned into an advantage.
For instance, since there are no mid-run save points (like waypoints) from levels 1 to 25, players would normally have to restart from level 1 every time.
But if a player lacks the gear and only manages to clear up to level 10, having 10 collected entry items (misc) for level 11 essentially gives them a save point that allows for at least 10 retries.

The-Honorable:
> > for lab 16-19 i think swapping it to the jail layout would be nice instead of re-using cata. unsure if we can use the act5 frozen cave areas or not for maze too.
> 
> We can use the ice areas. Using them isn't an issue in itself, except for the fact that they branch out into entrances, exits, and additional exits leading to other blocked spaces (like the Frozen River). While we could block those exits, it would likely affect other mazes and make things quite complicated. And since levels 1 through 4 are already using the jail layout, is there a particular reason to use it again for levels 16 through 19?
> 

if we are gonna have random exits that dont do anything we should avoid using them then. 

level 1-4 seem like they are using the swampy pits/flayer dungeon layout not jail

> > i do think to reduce the json amount we could easily just condense the level name strings to all just "Forsaken Labyrinth" instead of each level getting its own string.
> 
> Yes. However, I think it's better to leave the strings as they are while we're testing. That way, it'll be easier to identify exactly which level is causing the issue.
> 

sounds fine.

> > for lab boss level 15/20/25 i think we could just re-used lab 5 and remove the door vis for lab 25 or make 25 a different zone i am not leaning any particular way for that.
> 
> Yes, that sounds doable. The problem is that I haven't tried it yet, but I've heard that CTC Death can be a bit unstable. I believe there are cases where people spawn another monster when a boss dies and ensure that monster dies 100% of the time to mitigate this issue, but again, I can't say for sure since I haven't tested it. I'm also not certain if the portal missile will function correctly across all levels. If the portal to the next area fails to open even once when a boss dies, that would be a serious problem, so I think we need to give this some careful thought.

mmm yea needs tested im sure. would be the cleanest approach to avoid cube recipes.

The-Honorable:
> So, including the ones we haven't tested yet, there seem to be a total of three ways to transition from the boss stage to the next round.
> 
>     1. Both entrances and exits exist.
>        If we go this route, players will be able to skip that level one way or another.
>        A small map like Lab 5, in particular, could essentially be skipped immediately.
>        On the other hand, the development difficulty is the lowest. Of course, some map tiles in certain acts don't play well with each other when connected, but in most cases, they work just fine.
> 
>     2. portal is generated when the boss dies.
>        We eliminate the exit to the next floor and make it so that the only way to proceed is through the portal that appears when the boss dies. This would be the cleanest approach. However, as mentioned above, there seem to be some issues regarding Death CTC.
> 
>     3. boss drops a Misc item. That Misc item then serves as a cubing recipe.
>        Yes yes, I'm not a fan of cubing recipes being used this way either.
>        However, this would be the most stable method.
>        It allows us to physically separate each level (5 and 6, 10 and 11, 15 and 16, 20 and 21) while maintaining a 100% success rate.
>        The downside is that you have to go back to town after clearing the boss floor to open the portal.
>        Yet, this drawback could actually be turned into an advantage.
>        For instance, since there are no mid-run save points (like waypoints) from levels 1 to 25, players would normally have to restart from level 1 every time.
>        But if a player lacks the gear and only manages to clear up to level 10, having 10 collected entry items (misc) for level 11 essentially gives them a save point that allows for at least 10 retries.

1 is definitely the easier, but if with a little bit of testing 2 works that'd be cool. otherwise it can be a future effort down the line.

yinyin333333:
[test video link](https://www.youtube.com/watch?v=I8Yu4PdmpjU)

If you want to check from around Level 22, you can start watching from the 1-minute mark.
I've made two changes for testing compared to the PR I uploaded yesterday:

Changed the final level. (However, this alone didn't solve the issue.)

Skipped one level in the maze sequence. Instead of having 4 maze levels + 1 boss room, I changed it to 3 maze levels + 1 boss room. (Of course, Lab 15 and Lab 20 were already set to be skipped, but since our main concern was Level 25, I focused on that part.)

And this test seems to have confirmed it once again: D2R doesn't seem to like having four consecutive maze-type levels of a similar style. 
There might be other things I've missed, but for the sake of stability—and considering that standard D2R mazes (except for the Forgotten Tower) always cap out at three levels—a configuration of 3 maze levels + 1 boss room seems like the most appropriate setup.

yinyin333333:
[Test2 Record link](https://www.youtube.com/watch?v=RpZSa5R7BiU)

It’s just a process of continuing to build upon a very rough draft.
Here are the current changes.

Just as I tested earlier today, I have excluded one level from each set to have 3 maze levels + 1 boss room instead of 4+1. No crashes occur.
level 1,2,3 :  Act3 Dungeon
level 5 : Lab5
level 6,7,8 : Act1 Jail
level 10 : Lab10
level 11,12,13 : Act5 Nihl
level 15 : Temporary maps based on Harem (the sequence is subject to change).
level 16,17,18 : Act1 Cata
level 20 : Still keeping it skipped for now.
level 21,22,23 : Act5 Wolrdstone
level 25 : Temporary maps based on Cathedral (the sequence is subject to change).

image

yinyin333333:
I have cleaned up the level IDs  now, so the current levels are as follows.

01 (Act3 Dungeon)
02 (Act3 Dungeon)
03 (Act3 Dungeon)
04 (Act2 Lab5)
05 (Act 1 Jail)
06 (Act 1 Jail)
07 (Act 1 Jail)
08 (Act3 Lab10)
09 (Act5 Nihl's Temple)
10 (Act5 Nihl's Temple)
11 (Act5 Nihl's Temple)
12 (Act2 Harem Test temp map)
13 (Act1 Cata)
14 (Act1 Cata)
15 (Act1 Cata)
16 (Skip now)
17 (Act5 Wolrdstone)
18 (Act5 Wolrdstone)
19 (Act5 Wolrdstone)
20 (Act1 Test temp map)

And currently, there are no crashes across the 20 levels (actually 19, since I skipped one).
I think the other main contributors might choose to have it drop misc items for each next stage.
If we proceed that way, we just need to remove the exit to the next floor in each boss map (either by just disabling the exit function or removing the object entirely).

Since the boss maps all have a dark and gloomy feel, I considered changing the preset to an icy theme. However, I felt that a sudden transition to an ice background might seem out of place... and using the maze presets again is also a bit uncertain. I'll need to put more thought into this.
I could use the Act 2 mazes in addition to those from Act 1 and Act 3, but since the overall feel of the maze maps is similar—with only the tile colors being different—it doesn't seem like an urgent necessity right now.

yinyin333333:
Untitled-2
I didn't have any immediate ideas, so I connected a temporary map I made a while ago and left buried in a folder to the Level 16 slot. This is strictly temporary, and the order of maps, as well as additions or deletions, can happen at any time. (I'm just laying the groundwork.)
By adding one more temporary map, all levels (20 in total) now run without skips or crashes.
I don't think I'll have any more time to look into it this week due to real-life commitments. I'll have to check it out further next week.

yinyin333333:
I haven't had much time lately, but I'm taking a quick look at D2R today.
It turns out it wasn't an issue with consecutive mazes. I’ve successfully tested 25 levels without any crashes, and I'll update the draft within 10 minutes.

The-Honorable:
> I haven't had much time lately, but I'm taking a quick look at D2R today. It turns out it wasn't an issue with consecutive mazes. I’ve successfully tested 25 levels without any crashes, and I'll update the draft within 10 minutes.

oh that is really cool that you figured it out. interested to know what the fix is.

yinyin333333:
Just to be safe, I'll teleport around and do one more full sweep of the entire map. Give me a time few min.

yinyin333333:
I have rewritten the Lab with a 25-level configuration and confirmed with Ben that it runs without any crashes. Tasks such as mixing mazes, adding Act 2 mazes, or adding/removing specific boss maps will be handled at a later date.

yinyin333333:
> I have rewritten the Lab with a 25-level configuration and confirmed with Ben that it runs without any crashes. Tasks such as mixing mazes, adding Act 2 mazes, or adding/removing specific boss maps will be handled at a later date.

Okay. The current maze configurations aren't bad, but I’ve been checking out various options because having different-looking ones as spares would give other modders more variety to choose from. Hmm, it seems like there aren't any more, though.

1. A cave based on Act 1, Pit Level 1 – this one works. (Other caves have two exits, which causes the level to branch. If you're willing to accept that, other forms are also usable.)

2. Maggot Lair - Only part of it was possible. It’s probably best to assume it's practically non-functional, though I haven't looked into it further due to a lack of time.
Additionally, other maze types in Act 2 have several issues. The exit leading to the next level fails to spawn. I believe this is due to certain maps having hard-coded elements, and that seems to be the case here.
Because of this, it looks like there aren't any Act 2 mazes that can be used meaningfully besides the Tal Rasha ones. However, since the Tal Rasha maps are unidirectional, they don't seem to have much use other than creating a branching point in a different level that leads into them.

3. Durance of Hate Levels 1–2: In this case as well, the exit to the next level does not spawn. I suspect they might be usable as standalone levels, though.

4. Act 3 Sewers Level 1: Yes, the notorious map. I haven't tested this one. However, it cannot be used by simply cloning it. Since the files arranged in the maze have multiple entrances, the preceding map in the maze must also have multiple entrances like the vanilla version. Alternatively, the entire map would need to be created as a preset.
Since we are specifically discussing mazes, it cannot be used as-is.

5. Act 5 Ice Caves (Level 1): They all use the same maze. These can be utilized as mazes, but since each level has two exits, they will definitely create branching paths.
Of course, the ice background looks great because it’s blue, unlike the typical Catacombs or Worldstone Keep. However, to use an ice-themed maze, you would need to create branching points in other levels. Alternatively, there is a way to disable the maze warps in the existing vanilla maps, but that wouldn't be very meaningful.

So, unless We use presets or other manipulations, the ones currently on the list seem to be practically all of them, especially when considering stability within the maze structure.

The-Honorable:
> > I have rewritten the Lab with a 25-level configuration and confirmed with Ben that it runs without any crashes. Tasks such as mixing mazes, adding Act 2 mazes, or adding/removing specific boss maps will be handled at a later date.
> 
> Okay. The current maze configurations aren't bad, but I’ve been checking out various options because having different-looking ones as spares would give other modders more variety to choose from. Hmm, it seems like there aren't any more, though.
> 
> 1. A cave based on Act 1, Pit Level 1 – this one works. (Other caves have two exits, which causes the level to branch. If you're willing to accept that, other forms are also usable.)
> 2. Maggot Lair - Only part of it was possible. It’s probably best to assume it's practically non-functional, though I haven't looked into it further due to a lack of time.
>    Additionally, other maze types in Act 2 have several issues. The exit leading to the next level fails to spawn. I believe this is due to certain maps having hard-coded elements, and that seems to be the case here.
>    Because of this, it looks like there aren't any Act 2 mazes that can be used meaningfully besides the Tal Rasha ones. However, since the Tal Rasha maps are unidirectional, they don't seem to have much use other than creating a branching point in a different level that leads into them.
> 3. Durance of Hate Levels 1–2: In this case as well, the exit to the next level does not spawn. I suspect they might be usable as standalone levels, though.
> 4. Act 3 Sewers Level 1: Yes, the notorious map. I haven't tested this one. However, it cannot be used by simply cloning it. Since the files arranged in the maze have multiple entrances, the preceding map in the maze must also have multiple entrances like the vanilla version. Alternatively, the entire map would need to be created as a preset.
>    Since we are specifically discussing mazes, it cannot be used as-is.
> 5. Act 5 Ice Caves (Level 1): They all use the same maze. These can be utilized as mazes, but since each level has two exits, they will definitely create branching paths.
>    Of course, the ice background looks great because it’s blue, unlike the typical Catacombs or Worldstone Keep. However, to use an ice-themed maze, you would need to create branching points in other levels. Alternatively, there is a way to disable the maze warps in the existing vanilla maps, but that wouldn't be very meaningful.
> 
> So, unless We use presets or other manipulations, the ones currently on the list seem to be practically all of them, especially when considering stability within the maze structure.

for current config do you mean the maze type ordering?

and it might not be worth the hassle to deal with maze levels that have multiple entrances/exits unless you want to create empty little preset areas for the extra area which is like wrong way but here is a chest of loot for your time 🤣.

might just be better to ignore them though.

Delegus:
> create empty little preset areas for the extra area which is like wrong way but here is a chest of loot for your time 🤣.

miniboss?

The-Honorable:
> > create empty little preset areas for the extra area which is like wrong way but here is a chest of loot for your time 🤣.
> 
> miniboss?

would be an option too yea.

yinyin333333:
Yes. i mean current all maze.
It means there aren't many more mazes worth adding.
And yes, maze levels with multiple exits are a bit lackluster in that regard. (It's the same logic as moving from the Crystalline Passage to the Frozen River or other levels.)
It will just get unnecessarily complicated.
And the comment I left earlier was simply for testing.
It’s always better to check for sure, whether we end up using it or not. I was already aware of the hard-coded elements in some mazes where exits don't generate, but I didn't know that also applied to Act 2 and Act 3.

The-Honorable:
> Yes. i mean current all maze. It means there aren't many more mazes worth adding. And yes, maze levels with multiple exits are a bit lackluster in that regard. (It's the same logic as moving from the Crystalline Passage to the Frozen River or other levels.) It will just get unnecessarily complicated. And the comment I left earlier was simply for testing. It’s always better to check for sure, whether we end up using it or not. I was already aware of the hard-coded elements in some mazes where exits don't generate, but I didn't know that also applied to Act 2 and Act 3.

sounds good. did you have any thoughts on mixing the levels up so it's not same tile 4 times in a row? i know we talked about it, but i am unsure how mixed you were thinking.

like very mixed or sorta mixed?

dungeon
jail
nihl
cata
boss

or 
dungeon
dungeon
jail
jail
boss

yinyin333333:
Goodbye Azkaban, and welcome Chaos Sanctuary.
Azkaban doesn't really fit the concept of the current Lab. (Plus, it was originally just a temporary map used to test crash issues).
So, after thinking about which map to use, I decided to change it to Chaos Sanctuary.
Now, as Ben suggested, whether we mix up the order of the mazes or keep them separated is something we can decide based on everyone's feedback. (Of course, it's not up to me to decide. I'm just a spambot, and the other modders will make the decision.)

Oh, and for the entrance and exit of level 15 (Harem), I will eventually replace them with portals. It’s not a priority right now, though.

The-Honorable:
> Goodbye Azkaban, and welcome Chaos Sanctuary. Azkaban doesn't really fit the concept of the current Lab. (Plus, it was originally just a temporary map used to test crash issues). So, after thinking about which map to use, I decided to change it to Chaos Sanctuary. Now, as Ben suggested, whether we mix up the order of the mazes or keep them separated is something we can decide based on everyone's feedback. (Of course, it's not up to me to decide. I'm just a spambot, and the other modders will make the decision.)
> 
> Oh, and for the entrance and exit of level 15 (Harem), I will eventually replace them with portals. It’s not a priority right now, though.

thanks yin. and on that comment your input on this is valued too, dont sell yourself short.

data/local/lng/strings/levels.json
```diff
@@ -7172,5 +7172,1332 @@
         "ptBR": "Entrando no Cume Colossal",
         "ruRU": "Вершина колосса",
         "zhCN": "进入巨像之巅"
-    }
+    },
+    {
+        "id": 54000,
```
The-Honorable, line 7177: once we are done testing i think we condense these json strings into just 3 total instead of 3 for each floor as well since we are going to be hitting the json buffer limit very quickly otherwise.

yinyin333333: Yes, I understand what you mean. You’re saying we should have the four levels share the same string, and have the boss rooms share a single "boss room" string as well, right?

The-Honorable: maybe, i was even thinking it just stays 3 strings total and it wouldnt change. we can re-visit the idea later after we have a fully functional lab though.

yinyin333333:
gif1
Since there wasn't a boss map that matched the atmosphere of the Nihlathak area, I added a new map designed to feel like a long corridor stretching from the top right to the bottom left.
It replaces the existing Level 14 and can be used as a boss map.
I've placed it in Level 14 simply to avoid confusion, as I'm not sure how the maze settings will turn out; there’s no other reason behind it.
Since it’s quite long, I think the experience of fighting through waves of monsters before reaching the boss will offer a different kind of appeal compared to other Labs.

yinyin333333:
While eating breakfast, I thought about including a few Cave Maze levels for testing purposes, even if they aren't used in the final version. This led me to reconsider Death-CTC, and I realized a potential issue that arises when applying CTC to the current structure.


Looking at the diagram, this is the current structure.
Several mazes and one boss room; they all have entrances and exits, so naturally, there are no issues.


And this diagram illustrates the scenario where Death-CTC is implemented, requiring the boss to be killed before progressing to the next floor.

The absence of an exit in the boss room implies that the subsequent floor must also lack an entrance. Therefore, even if we manage to implement Death-CTC with 100% stability, the room following the boss room must be a preset room with its entrance removed. Given our current structure of "Maze + Boss Room," this would essentially become a dummy preset room—literally a room just to pass through.

This would need to be added for every boss encounter. It would be a meaningless transition room, though it could technically still be used as a hunting ground. However, this might make the overall configuration quite complicated.

yinyin333333:
adding the ice-themed maps for testing purposes.
The current map configuration is as follows, and we won't be using every single map.

Forsaken 01 (Act3 Dungeon)
Forsaken 02 (Act3 Dungeon)
Forsaken 03 (Act3 Dungeon)
Forsaken 03 (Act3 Dungeon)
Forsaken 05 (Act2 Lab5) --- boss map
Forsaken 06 (Act 1 Jail)
Forsaken 07 (Act 1 Jail)
Forsaken 08 (Act 1 Jail)
Forsaken 09 (Act 1 Jail)
Forsaken 10 (Act3 Lab10) --- boss map
Forsaken 11 (Act5 Nihl's Temple)
Forsaken 12 (Act5 Nihl's Temple)
Forsaken 13 (Act5 Nihl's Temple)
Forsaken 14 (Act5 Nihl's Temple) --- boss map
Forsaken 15 (Act2 Harem) --- boss map
Forsaken 16 (Act1 Cata)
Forsaken 17 (Act1 Cata)
Forsaken 18 (Act1 Cata)
Forsaken 19 (Act1 Cata)
Forsaken 20 (Act1 Cathedral) --- boss map
Forsaken 21 (Act5 Wolrdstone)
Forsaken 22 (Act5 Wolrdstone)
Forsaken 23 (Act5 Wolrdstone)
Forsaken 24 (Act5 Wolrdstone)
Pit Level 1
Pit Level 1
Pit Level 1
Pit Level 1
Crystalline Passage
Drifter Cavern
Crystalline Passage
Drifter Cavern
Crystalline Passage
Drifter Cavern
Forsaken 25 (Act4 Chaos) --- boss map

yinyin333333:
original levels have been backed up, and now it is time to shuffle the maze and establish the proper sequence.
I tested every map across all 25 levels by revealing the entire field of vision, and as expected, there were no crashes or issues.
Below is the current level structure.

```
├── Forsaken 01 (Act1 Pit Cave)
│
├── Forsaken 02 (Act1 Pit Cave)
│
├── Forsaken 03 (Act1 Inner Cloister)
│
├── Forsaken 04 (Act1 Outer Cloister)
│
├── Forsaken 05 (Act2 Lab5) ---Boss Map
│
├── Forsaken 06 (Act3 Dungeon)
│
├── Forsaken 07 (Act3 Swampy Pit)
│
├── Forsaken 08 (Act3 Dungeon)
│
├── Forsaken 09 (Act3 Flayer Dungeon)
│
├── Forsaken 10 (Act3 Lab10) ---Boss Map
│
├── Forsaken 11 (Act5 Halls of Anguish)
│
├── Forsaken 12 (Act5 Halls of Pain)
│
├── Forsaken 13 (Act5 Halls of Anguish)
│
├── Forsaken 14 (Act5 Nihl's Long Map) ---Semi Boss Map?
│
├── Forsaken 15 (Act2 Harem) ---Boss Map
│
├── Forsaken 16 (Act 1 Jail)
│
├── Forsaken 17 (Act1 Cata)
│
├── Forsaken 18 (Act 1 Jail)
│
├── Forsaken 19 (Act1 Cata)
│
├── Forsaken 20 (Act1 Cathedral) ---Boss Map
│
├── Forsaken 21 (Act5 Ice Cave)
│   └── Forsaken 21 (Act5 Ice Mini Dungeon)
│
├── Forsaken 22 (Act5 Worldstone)
│
├── Forsaken 23 (Act5 Ice Cave)
│   └── Forsaken 23 (Act5 Ice Mini Dungeon)
│
├── Forsaken 24 (Act5 Worldstone)
│
└── Forsaken 25 (Act4 Chaos) ---Final Boss Map
```

PR #1179, opened 2026-04-14
fix: herald self damage problems

Delegus:


The-Honorable (approved):
harold aint so frosty now

PR #1189, opened 2026-04-18
Grandfather sprite spruced up, Eth removed, ED added

GildyBoye:
Eth removed, gave flat ED% to add more variance

image
image


The-Honorable (approved):
seems fine. cool sprite

PR #1191, opened 2026-04-18
quivers work in loot filter properly

The-Honorable:
both quiver types show up under weapons and are properly filter-able now. can always re-adjust this to be under accessories or armor later if that is preferred.


Delegus (approved):
hoo ray

PR #1194, opened 2026-04-21
fix: Swap TZ sound change

Delegus:
Apparently 0 size .flac triggers the nono sensor. Swapping tweak to sounds.txt


The-Honorable (approved):
so quiet

PR #1198, opened 2026-04-24
Holy Auras rebalance

strangemolars:
Reasoning:

Numerically, you could always reach absurd numbers with Holy-Aura stacking ever since I started playing due to the mastery double-dip these skills are suceptible to, and how much more you can stack those stats here when compared to Vanilla. 

What happened in the past is that these build competed directly at the end game with "old" Vengeance, that was a very similar skill that also reached numbers over the 100ks even easier; and that these auras were limited to using Zeal as a vehicle of delivering damage.

Now, both of those things are not the case anymore. Vengeance is now, besides a complete build on itself, also the best vehicle to abuse holy auras. On itself, Vengeance is a slower attack than Zeal, but since it is an aoe attack on its own merit, it magnify the auras effect both in that it multiplies its actual tooltip numbers (that are already prone to "multiple multipliers") by +78%, and now makes the build also benefit from splash charm double-pinging the added aura on attack damage. The aura damage added on attack is added to the splash charm radius, then also to the main vengeance target 100%, then it is added 26% as each of the 3 elements on the Vengeance radius. Thats 2.78 times the already high on attack damage listed on the main target, and 1.78 times the listed damage on the aoe. Remember this when considering the numbers I'll post on the screenshots.

Changes made:

Holy Fire:
Base tick multiplier added to melee attack from x6 to x4;
Min and max damage curves adjusted half point down on curves 2 and 3, and half point up on curve 5. Kept the same at curves 1 and 4.

Holy Freeze:
Base tick multiplier added to melee attack from x5 to x4;
Min and Max starting damage adjusted 1 point down at the first lvl (omg the nerf);
Synergies adjusted from 15% and 7% to 14% and 6%.

Holy shock:
Base tick multiplier added to melee attack from x6 to x4.

Testing results:

All auras attack damage brough some more in line. Due to how overtuned these auras are baseline, their usage as "leveling tools" that a lot of people would rely on for is basically unaffected. Holy fire was most tested on this regard, since it is the most used one for this purpouse. I felt that the change in damage was not noticeable during normal. During late stages of nightmare it was noticeable was little bit more noticeable, but still very strong for the content. While swapping around with zeal and vengeance to test the impact on both enablers, this was obviously felted a lot more on Zeal, but that is due to the "fair" nature of that skill. Vengeance felt much less the damage nerf since its an overall stronger and more synergistic skill mechanically to Holy-Auras melee added damage.

Holy Fire tick damage was very slightly buffed. Pushed on a 4 tick stacker that I used to beat hell (regular HoJ, Flame, Dragon, Dragonscale Holy Fire build), it was nice but hardly something that make or break chars. On a single pulse high geared chareacter, the difference on the aura pulse was from 9.5k to around 11k damage every pulse. As single pulse, still relative weak damage.

Tested the changes by running HTZ for about 2 hours on Prodigy's endgame Holy Fire self-wielded char both on x6 and the x4 changed chars. Very geared chars are expected to basically one-shot density at this point, and the build still does that fine. The damage nerf is felted on chonkier engagement like act bosses, but it feels more in line. For me everything could be nerfed more, but I'm trying to not go overboard here.

Tested the effects of damage on Holy Shock on a pretty geared character too. I ran my Lightning Ranger char for a couple of days on HTZ farming stuff for my merc. Here the impact is even harder to judge exactly due to the huge variance nature of Holy Shock aura, but the nerf also didn't felt drastic at all.

Tested how Holy Freeze felt with the "relative buff" it had compared to the other 2 auras. While Fire and Shock got chopped from x6 to x4, Freeze only got chopped from x5 to x4, which brought it closer to added damage on attack when compared in particular to Holy Fire. The Holy Fire and Freeze pulse damage never were that much different to one another in raw curves, but the difference in the melee added damage, synergies and most importantly gear are the things that compounded made Holy Fire a lot more viable as a damage source than Freeze. My intention is not to make Holy Freeze too close to Holy Fire, otherwise it is just a better aura any way you can slice it. I kept looking at numbers at different levels, and trying some equip gears, and Holy Freeze was still just kept at a safe enough distance from Holy Fire so it still feels like a very aptly costed choice between damage and safe to use Freeze. Therefore, I just did a very slight synergy decrease just in case, and kept everything else as is.

The last thing I did was testing, just to be extra safe, was a 6-way pulse stacker for holy freeze. It consisted of Sine chestpiece, 2x Chiad rings, Meph gloves, Doom one hander and Tools of Vindication Shield. That added up to a measly lvl 45 Holy Freeze aura, pinging 6 times due to how on-equip aura interactions work. It was still weaker than a 4x pulse lvl 62 holy fire, so I am convinced its not also a problem. It also doesnt have much to do with what I'm rebalancing here, but I did it just in case.

Screenshots of in-game impact:

A holy Fire lvl 25 with sinergies and nothing else, before and after nerf.
tt252
tt251

A holy fire + holy freeze double-stacker pally I used to beat the game with the nerfs. Gear was the usual 4-stacker HF items (HoJ, Dragon helm, Flame Chest, Dragonscale shield. No enchants, no corrupts, Um runes socketed on shield. No extra %mastery from any source. )
1HAX6
1HAX4

An endgame geared paladin (Prodigy's one), that has very good items - most of what I think someone that invest some time in the char would get - but its not 100% optimally build. I'm prettu sure you can push these over 100K without any extra multiple roll absurdness because I did have a pimped out double stacker (which this one isnt) reaching the 110k range (without respec bug) in 2.1.
2HAX6
2HAX4

The tooltip damage of the above char. Notice that that char has 161% extra fire damage, which multiplies again this now fixed tooltip damage when adding to the attack:
2HAX6tt
2HAX4tt

I also have my lightning ranger char that I can take screenshots of, but its a bit more of a laborious process because you can only actually gauge the effect of the nerfs when I do weird partial gearings, as that char is prone to the super weird lightning damage in excess of 63.5k bug I crashed out on in the contributor chat some weeks ago. Enough to say that the theoretical damage the x6 version should be doing was 127.7K added light damage, and the x4 is about 92K. On the char tooltip, these were 95.5k and 92k respectively, which is just complete "what the fuck is going on on this game". I was able to push that x4 character to around 105k damage with the gear I had available and the 63.5k damage thing didnt show up again untill that point. No reason why changing the multiplier from x6 to x4 would fix that (display?) bug, but it did. D2 as we love it.

TL;DR:
I think if anything the nerfs I'm proposing could not even be pushing enough; but I'm not trying to kill these specs nor anything. These are some of the chars I enjoy playing the most, and I just want them to be a bit better balanced in the late-game while still maintaining the use they have to the avarage players that just use them to level. Also, there is the consideration of the Splash Charm being, eventually, removed and that will affect the damage in a negative way too.


Delegus (approved):
I didn't read this, I'm sure it's fine, we're reworking stuff anyways. Surely.

PR #1199, opened 2026-04-24
adjust gamble cost to 1 for quivers to fix their price issue

The-Honorable:


Delegus (approved):
Going broke buying ammo

PR #1203, opened 2026-04-26
adjust starting items

The-Honorable:
- All classes now start with ID Tome and TP Tome instead of 1 scroll. 
  - These tomes are full since using them in any recipe also refills them   
- Sorc now starts with an eagle orb + buckler
- Warlock now starts with a buckler
> Unsure why necro and druid had a bow. While Necro has a bow skill it is not available at level 1. Druid has no bow skills.
- Necro now starts with a buckler
  - Also removed starting bow/quiver
- Druid no longer starts with a bow/quiver

## Items
- Buckler strength requirement from 12 to 10 (for Sorc basically)

## Affixes
- Lowest level large charm prefix and suffix ilvl to spawn changed to 1 to prevent very niche case crashes


Delegus:
image

Delegus (approved):
Left you a note in the comment.

PR #1206, opened 2026-04-27
Blessed Hammer build synergy swaps

strangemolars:
Wont get too much into it because it was memed to death on the last one. Still a Shooby post though. xD

Changes made:
Swapped Blessed Hammer synergies from Vigor and Blessed Aim (2 "WTF" synergies imo) to Holy Bolt and Prayer.

Reasoning:

Holy bolt - to be a tool in thight spaces. Hammers inherit playability problem remains, but this addition can alleviate it a little. Since it doesnt have a second synergy for Holy Bolt like the Fist spec can have, and the skill doesnt benefit from Concentration Aura, the Holy Bolts are weaker and you will prefer to use Hammers even on bosses when you have the space to do it. But at least you have mid-damage Bolts available when you are tucked in places like maggot lair, flayer dungeons, and the likes. Specially helpful while leveling - if someone is masochist to do it as hammers.

Prayer - there arent that many choices on what to pick to still force Hammers to be different build than FoH, specially with the overlap of HB, and still give the build some interesting dimension. The way I found was prayer, because then the player has the option, if they want, to also use gear like for example Celestial Unicorn and Knight's Vigil to have passive Prayer ping going even in their regular Hammer gear (single and double ticks available). This felt fun in the gameplay I'm doing, and felt like a good fun perk to have over FoH since Hammers demands 100 point hard investment with basically no flexibility; and it didn't felt overpowered at all.

I didnt go into any extent to alter hammers damage itself. The build feels a bit weaker compared to FoH spread, but I think it does completely acceptable numbers.

Extra stuff to consider: HB doubling as Hammers and FoH synergy could lead to some extra possibilities of builds using stuff that CTC proc hammer and FoH. If magic mastery is made acessible and able to be amassed on items geared towards this, then you could even have the Zeal+Sacrifice synergy of 25 magic damage to Holy Shield be relevant and along with the procs it could lead to some cool "Magic Zealer" Hammer CTC builds.


Cantrip-commits-crimes:
Why not Concentration / Holy Bolt, if you don't wanna just make the 3 Holy skills all work with each other? Concentration has unique interactions with Hammers atm, so it fits.

strangemolars:
> Why not Concentration / Holy Bolt, if you don't wanna just make the 3 Holy skills all work with each other? Concentration has unique interactions with Hammers atm, so it fits.

Exactly, its already a "fake" synergy. So why make a synergy be even more of a synergy? Double-dipping synergies is kinda dumb, specially when it is on the same axis (it would be damage synergy + damage increasing effect). This would also change the build from its current point spread to be 25 less than it is now, which would be a buff that I didn't get to test nor I think it necessarily need. Even if the intention was reducing the build spread, then making the HB synergy be double and not have Concentration be a formal synergy would still be a better choice imo.

The intention is just giving it a bit more pzazz, not hard rebalance it.

Bakelt:
Would not double dip. Having an offensive aura as a synergy is just going to cause number issues and unnecessary math for balancing.

I'm already removing Fanaticism from Sacrifice for the same reason.

strangemolars:
> Would not double dip. Having an offensive aura as a synergy is just going to cause number issues and unnecessary math for balancing.

I didnt meant "double dip" as in what happens on Sorcs and Paladins and etc, I just meant as in that Concentration already gives Hammers damage as its baseline effect, and making it a formal synergy that would also just give it damage would be redundant. The same spell would be doing 2 times the same thing... which feels weird.

The-Honorable:
initial look, it seems fine. dunno if you took a look at any uniques/runewords for hammers at all

at some point the skill curve of the magic skills needs a proper review in order to re-itemize more magic mastery onto items too.

Delegus:
Original synergies look like they were just placeholders that got shrugged and fugged 🤣

Bakelt (approved):
LGTM

PR #1207, opened 2026-04-28
Happy Maggot Lair

yinyin333333:
The Maggot Lair is now a place full of happiness, not frustration.
The tunnels are being widened.
Also, the path layouts have been simplified, making it much more convenient regardless of the wider passages.


yinyin333333:
As you can see from the shape of the minimap, it becomes simpler. It doesn't mean the maze itself becomes simpler, but since I reduced unnecessary bends and curves, it will be much more convenient to walk around.



We now have enough space to pitch a tent and camp right in front of the Maggot Queen. (Of course, Druids aren't allowed. No animals, including wolves and bears, are permitted entrance.)

But this PR is just a suggestion, and I don't like the Maggot Lair myself.
Maybe I should consider blowing up the entrance with dynamite and just blocking it off entirely.

Bakelt (approved):
LGTM - minor issues with minimap if you've already explored the maggot lair but nothing we can do about that

yinyin333333:
> LGTM - minor issues with minimap if you've already explored the maggot lair but nothing we can do about that

Yes. The only way to fix that is to reset the map by changing the difficulty or using the reset offline map feature.
Alternatively, we could change the offset so it doesn't overlap with the existing minimap, but altering the characteristics of the vanilla map just because of the minimap feels like the wrong approach.

PR #1209, opened 2026-04-29
Plugin Support for Aura Spiritz

Cantrip-commits-crimes:
Adding placeholder monsters & skills to swap in for auras with a plugin.


Cantrip-commits-crimes:
This isn't for merging, this is prep work for a plugin.

PR #1210, opened 2026-04-30
KOR translation: Fix invalid `x per level y` translations

d3m3vilurr:
only check sentries of sin


d3m3vilurr:
i already shared it on discord, sin strings have to change `per base lv %d`. when change the eng still, i'll update this again.

Delegus (dismissed):

Delegus (approved):

PR #1212, opened 2026-05-01
feat: Fix missing UI.json strings added in last SnowCo update

Delegus:


Cantrip-commits-crimes:
Looks like it won't catch fire.

Cantrip-commits-crimes (approved):
lgtm

PR #1213, opened 2026-05-02
Add Statues and Shards to no-DLC for funsies

### LINKED ISSUE #1202: Add controller settings for trangs meteor and firewall
Title

### LINKED ISSUE #1204: Add TC Entries for Ancient Statues and Shards to Base
Title

Delegus:
Resolves: #1204 
Resolves: #1202 

Allows for shard and statues to drop and be used and the uber ancients to be fought on no-DLC characters.
Fix controller support for Trang set oSkills


yinyin333333:
I know, we know there are no issues because I previously tested and confirmed that Uber Ancients works fine in non-DLC. It's a welcome update for non-DLC users.

The-Honorable (approved):

PR #1215, opened 2026-05-02
Necromancer bone spells change / overhaul

ZiRxXeNoN:
Adds Wraith Malediction / Bone Storm, tweaks Bone Armor and synergies and some other necromancer skills to accomodate the change


The-Honorable:
before anything you need to rebase off the next branch. it looks like you pasted old code a lot of places are are reverting a lot of changes other people have made recently. its generally better to not paste whole files and instead make your changes directly in the files.

Delegus (changes requested):
Get this up to date with next, then apply your changes, this is outdated.

Delegus:
These are to be made into stand alone mods/plugins. As they are they don't fit and this was the direction chosen after that was explained.

PR #1216, opened 2026-05-03
Arctic Blast Rework

Bakelt:
### Arctic Blast
- Reworked. Now creates an area that does cold damage over time against targets caught in the area

### Twister
- Missile speed increased
- Minor damage nerf

### Tornado
- Missile speed increased
- Missile radius increased
- Minor damage nerf

### Hurricane
- Now gains 3 seconds per level (up from 1)
- Now gains 3 seconds per synergy level (up from 2)

### Armageddon
- Now gains 3 seconds per level (up from 1)
- Now gains 3 seconds per synergy level (up from 2)


yinyin333333:
2026-05-0312-57-47-ezgif com-video-to-gif-converter

It looks cool and refreshing, good.
Since Druids are characters that originally lacked this kind of freshness, I think it suits him well.

Delegus (approved):

PR #1217, opened 2026-05-03
fix summoner AI/skills/nerf hp slightly, enable regen because i'm evil idk

The-Honorable:
## Monster Changes
- Adjusted his AI so he will actually cast his skill2 (fireball)
- Fixed his teleport animation as it was preventing him from actually teleporting and putting him into hit recovery instead
- bumped his movement speed up slightly
- lowered his HP slightly as he will be a bit more dangerous in nm/hell
- Enabled 1 regen, can be changed later if giving him regen is too mich

### Shaman Fireball
- Projectile count changed to 1 proj until level 4 which it goes to 3 proj, until level 8 it goes to 5 missiles

- Basically the same as shamans fireball. can be reviewed again later


yinyin333333:
2026-05-0313-17-37-ezgif com-optimize

frequency of Fire Ball and Glacial Spike has become similar, and I confirmed that he fires Multi-Shot as well. (The attached gif is from Nightmare difficulty.) Fire Wall seems to be the next most frequent, and Teleport is used at a much lower rate than in the previous commit, which I think is fine.

Bakelt (approved):
LGTM

PR #1221, opened 2026-05-04
Fully Magic Fist of the Heavens

Cantrip-commits-crimes:
Main hit is Magic damage, Synergy with Blessed Hammer instead of Holy Shock.

Strings corrected accordingly.


ProdigyWTF:
But where's holy

PR #1225, opened 2026-05-06
drain/repl string changes because mechanics are hard to understand i guess

The-Honorable:
changes drain and replenish strings to indicate that the mechanic works continuously but the number shown is functionally /10 per second

image

image


Delegus (approved):
🖍️ 🥫

PR #1226, opened 2026-05-06
small necro tweaks

The-Honorable:
goal: tweak some bone necro things to rework a useless skill and minor adjustments to the rest

- teeth missile speed up 25%
> teeth is quite capable at clearing the screen, especially with a bit of a missile velocity boost.
- review giving bone spirit some sort of AOE explosion/nova
> bone spirit hits much harder than bone spear, often times near double its damage making it great when there is a single enemy nearby, but less great when multiple enemies exist. Adding some sort of nova would make it a bit more reliable when trying to tackle tough enemies that are surrounded by allies
- teeth/bone spear/bone spirit synergy to bone wall/prison removed
- bone prison to be reworked into more of an ultimate skill unlocked at level 30 with a requirement of bone spirit instead of bone spear
  - current thought is something that has a cooldown of ~10 seconds that possibly targets a corpse and explodes releases chaining bone spirits/bone shard like projectiles that deals massive damage
- the new bone prison skill would have a synergy with teeth/bone spear/bone spirit/energy

- potentially give bone armor an energy synergy, adjust it's current numbers and remove the bone prison synergy
- bone wall, remove bone prison synergy. is fine functionally and is one of the best defensive tools in the game.

its not done so settle your acid pikachu dong arms


Delegus:
I've said before why bother changing these synergies just to remove them, I know BakeIt has said it as well, do you plan to replace wall and prison in some manner? Currently the only saving grace for these two defensive spells is at least they provide a synergy. You don't even play necromancer, you are just randomly looking at numbers and flailing.

strangemolars:
This one kinda confuses me regarding the synergies too. They are the only thing that make Bone Armor usable for bone necros, and if Teeth is ever made interesting, it will be an interesting choice to have picking between more offensive spells or defensive ones on your build. Having a bunch of synergies in the way this one have is only a consideration for people running 2 or more skills per level, but in that case, whatever. If they run into overflow thats a problem of their own doing. Removing these skills from the synergy slab doesnt impact damage at all on the current balance, only utility.

Thats not to say that whenever/if bone prison and wall get replaced with some other stuff, this cant change.

I like the funny belt remake though.

The-Honorable:
will open a new pr on newer files

PR #1229, opened 2026-05-07
Zenith

crypto1819:
Adds the Zenith skill to the paladin


crypto1819:
Not quite ready yet — this is still pretty bare-bones.
I'd like some additional guidance before moving forward, specifically:

Should this replace an existing skill, or is it meant to be a brand new addition?
Any preference on which skill it should replace (if applicable)?

I don't have a sprite for it yet, but I'm happy to create one. Let me know how you'd like to proceed!

yinyin333333:
I played it for a bit short time.





The clarity of the overlay created by the two-handed sword slam looks good. However, the color doesn't seem to match the Firestorm-like visuals that follow on the ground.

Is this intended to be used as a kind of ultimate? (I haven't read the skill description.)
Given the flashy nature of the skill, it might be a good idea to set it as an oskill and grant it to specific unique items. The Paladin is fine, but it's also fun using that skill with a Sorceress. (Of course, this is just my small opinion.)

yinyin333333:
image

> clarity of the overlay created by the two-handed sword slam looks good. However, the color doesn't seem to match the Firestorm-like visuals that follow on the ground.

It seems the translation got reversed. Stupid yin.
The vivid color of the overlay is good, but the color of the flames on the ground feels out of place compared to the color of the overlay.

crypto1819:
Hmm, seems like the fire isn't getting the tint. I'll look into that.

Cantrip-commits-crimes:
IIRC, Zeal was gonna be replaced or reworked now that Sacrifice is actually good, this could be an option for that

@Bakelt

Delegus:
> IIRC

You don't.

Bakelt:
We would likely replace an unused aura and have 11 skills in combat tree.

Bakelt:
Skill looks clean based on the above. The fire looks like it persists too long. Would drastically reduce the Range column (# of frames missile is active) on both missiles until the overall range of the missiles is more in line with other area attacks.

Currently it looks like both the missile maker and the fire itself are dealing full Weapon Damage per frame so that will obviously need to be changed. I would either
1) Make zenith_firestormmaker deal a % of weapon damage using hitshift/SrcDam (see Shockwave) and have zenith_firestorm be visual only. Have the firestorm duration real short so it is just a short flash of fire. Will result in a snappy AoE clear like Tectonic Slam
2) Make zenith_firestormmaker deal 0 damage and give zenith_firestorm it's own scaling in missiles.txt to deal damage over time like firestorm does. Can use firestorm curve as a base, add ApplyMastery flag, NHD=0, and that would mostly cover it.

Personally, I'd also change the skill req from "mele" to "slam" to limit it to the 2h weapons to enforce 2h Pala identity.

yinyin333333:
It seems crypto is no longer active. Regardless of this skill's performance, I really like its visuals and think hardly anyone would dislike it.
Does anyone have other plans for this?

yinyin333333:
This PR seemed forgotten, so I made a few adjustments. (I'm not saying we should do it exactly like this, it's just a boost for something that got forgotten. I figured this would make Del or Bakelt show up.)
Since the rule about every class having the same number of skills no longer applies after Rotw, I just set it to pal.
Also, we'll need to create the sprites anyway, so I just set it up like that as a placeholder. (I don't have any strong opinions on skill damage, level requirements, or skill prerequisites—I just set them like that for now.)
Personally, it bugs me a bit that the ground fire's color doesn't match the overlay, but since it disappears so quickly anyway, maybe we don't need to make a new one.

image

PR #1100, opened 2026-03-10
Act 3 Merc Changes

Bakelt:
### Leap Attack
- Fixed leap attack not dealing AoE damage
### Tectonic Slam
- Made tectonic slam projectile faster

## Mercenaries
- All Act 3 mercenaries now have their own elemental enchant which also gives a small amount of resistance piercing
- This enchant does not stack with sorc's enchant 
- Each A3 merc now primarily casts fireball/glacial spike/chain lightning and occasionally a spell unique to their element that deal increased damage.
   - Icicle Barrage: A series of frost bolts form and are sent out in a wave
   - Fire Twisters: Several fire twisters dealing damage in an area
   - Lightning Nova: A lightning projectile is sent out that explodes into a small nova
- Improved A3 merc damage scaling for all skills
- Improved A3 merc skill level scaling for all skills. All start at level1 and will gain +1 level every other level (max skill level 43 at level100 before gear)

PR #1231, opened 2026-05-07
Mosiac Is Kill and BossSkills

Bakelt:
Just a test branch to throw some shit at

BigBoom / Safenova
- Creates a deadly nova outside a safe spot. Safe spot is currently marked by apocalypse but should change

Boss Armageddon
- Creates sigils around the user that deal damage after they expires (2s)
- Can easily adjust the frequency of sigils as well as how far away from the caster they can appear. Going below missile.Radius 4 causes bugs

## Assassin
This is an early look at a possible change for the MA skills. Goal is to remove Mosaic as it is problematic while still having a combo feel to the assassin MA skills. This is still heavily under development, buggy, and susceptible to change

Fists of Fire = FoF
Claws of Thunder = CoT
Blades of Ice = BoI
Phoenix Strike = PS
FoF1/2/3 = FoF Stage 1 / FoF Stage 2 / FoF Stage 3

<<Known Issues/Bugs>>
- Skills may default to the first missile when the player is sufficiently slowed. Burst of Speed helps negate this early before the player has access to Cannot Be Frozen
- Skills may repeat the current missile if the player's attack speed is too fast. There is currently a workaround blocking this but I have not tested all IAS breakpoints
- The skills do not want to handle their %damagepercent scaling properly for the phys portion and are likely doing less damage than what the character sheet implies

<<Changes>>
Skills use an internal counter that increments with EVERY skill/attack. This creates a gameplay loop similar to D3 Monk primary attacks where casting Fists of Fire (FoF) > FoF > FoF will result in FoF Stage 1 > FoF2 > FoF3. 

Casting FoF > BoI > CoT will result in FoF1 > BoI2 > CoT3

-- Fists of Fire (FoF) -- 
	Stage One: Increased flat fire damage to the target and one nearby target
	Stage Two: Fire nova
	Stage Three: 8-way fire damage over time

-- Claws of Thunder (CoT) --
	Stage One: Increased flat lightning damage to the target and one nearby target
	Stage Two: Lightning nova
	Stage Three: A scattering of small lightning novas

-- Blades of Ice (BoI) --
	Stage One: Increased flat cold damage to the target and one nearby target
	Stage Two: Cold nova
	Stage Three: 5 piercing frost bolts at nearby enemies

-- Phoenix Strike (PS) --
	Stage One: Meteor (same as vanilla PS1)
	Stage Two: 3-way lightning
	Stage Three: cold tornado that hits nearby enemies 2-3 times

Remaining skills are unchanged and still operate under the old system while designs are made for new skills. Tiger Strike will likely remain a phys combo skill. Dragon Flight will remain unchanged as primarily a utility skill. Cobra Strike, Dragon Talon, Dragon Tail, and Dragon Claw are undecided.


data/local/lng/strings/skills.json
```diff
@@ -34155,19 +34155,19 @@
...
-        "ptBR": "Carga 2 - Raio de dano ígneo: %d metros",
-        "ruRU": "2 заряда – радиус поражения уроном от огня: %d м",
-        "zhCN": "2次充能 - 火焰伤害范围: %d码"
+        "enUS": "Charge 2 - Fire Nova Damage: %d-%d",
+        "zhTW": "兩次充能－火焰新星傷害：%d - %d 點",
+        "deDE": "Ladung 2 - Nova-Schaden: %d-%d",
+        "esES": "Carga 2: daño de nova: %d-%d p.",
+        "frFR": "Charge 2 - Dégâts de Nova : %d-%d",
+        "itIT": "2 cariche - danni Nova: %d-%d",
+        "koKR": "충전 2 - 화염파 피해: %d-%d",
```
d3m3vilurr: i thought we told it when #1100 

but it's not, we didn't decide translation of the `Fire Nova`

it's same as (electic) nova but fire version? (or it new plan to add new fire version nova skill?)
if that just give a name `화염 노바`

or if it's same as ring of fire of warlock, how about change name to this?

data/local/lng/strings/skills.json
```diff
@@ -34206,36 +34206,36 @@
...
-        "ptBR": "Carga 2 - Raio de dano gélido: %d metros",
-        "ruRU": "2 заряда – радиус поражения уроном от льда: %d м",
-        "zhCN": "2次充能 - 冰霜伤害范围: %d码"
+        "enUS": "Charge 2 - Frost Nova Damage: %d-%d",
+        "zhTW": "兩次充能－冰霜新星：%d - %d 點",
+        "deDE": "Ladung 2 - Frostnova: %d-%d",
+        "esES": "Carga 2: Nova de escarcha: %d-%d p.",
+        "frFR": "Charge 2 - Nova de givre : %d-%d",
+        "itIT": "2 cariche - Nova Gelida: %d-%d",
+        "koKR": "충전 2 - 서릿발: %d-%d",
```
d3m3vilurr:
```suggestion
        "koKR": "충전 2 - 서릿발 피해: %d-%d",
```

data/local/lng/strings/skills.json
Bakelt: The second attack in each combo is an elemental nova of the same element. So basically a clone of the existing sorc Nova and Frost Nova with slightly different visuals (and a clone of fire claw splash for a fire nova)

d3m3vilurr: well... can this make a change to add new skill to sorc in future?
if that, I like to define `fire nova` then give `화염 노바`
(wow uses fire nova to `불꽃 회오리` imo this name confuse to tornado)

i think it's another origin sin of the bliz. idk, why they give a name `ring of fire` to fire nova skill of the warlock. :/

PR #1232, opened 2026-05-08
Golem Damage and Chasm Break Lag Updates

### LINKED ISSUE #1188: Chasm Break Lag
Noticeable in other acts but particularly bad in Act 3 with spiders, Chasm Break is causing the game to drop from 120 FPS to 10 while hitting large packs of spiders, and in other areas from 120 to 40 on packs of enemies. Need to investigate the behind the scenes as well as the particles and see if I can determine the cause.

### LINKED ISSUE #1192: Clay / Iron / Blood Golem Splash doesn't work
Clay / Blood use item_normaldamage (doesn't splash), Iron just doesn't have Splash at all.

(Yeah, I tried to fix it, Del, gave me a headache)

Delegus:
Resolves: #1192 
Resolves: #1188 

### Carnage
- Reduced damage per projectile to 25% from 50%
    >This skill is still wildly overperforming due to the original change that tightened the projectile spread, if this doesn't bring it back in line then I'll start working the opposite direction, boosting damage but making it less reliable.

### Chasm Break
- Reduced erroneous visual and missile counts causing lag and crashing

### Sacrifice
- Fixed a bug where the splash was using AR

### Shield Throw
- Bug found where Shield Throw was not using AR. Since this is paired with Smite which doesn't use AR, this behavior will be kept for now.
- Character sheet will no longer show AR since it always hits

### Summon Resist
- Completely gutted and doesn't do anything in prep for replacement

### Skeleton Mastery & Golem Mastery
- These two skills inherited the Summon Resist passive. Bonus levels * 3 and caps at 84. They will not stack, higher skill level wins and the other disables, in case someone is building pure Skeletons or Golems without mastery in one
    >This summon resist bonus is still slower than druid pets.

### Golem Mastery
- Now applies damage 5% synergy per base level to all Golems damage curve  instead of flat damage

### Clay Golem
- Rescaled to use damage curve, does not self grow  with damage % until level 25 or higher due to early acquisition
- Now receives second golem at level 10 (was 20)

### Blood Golem
- Rescaled to use damage curve, self scaling locked at 18+
- Leech Scaling is now 1% baseline with 1% per level, still only provides 1% of self-leech to necromancer
    >This guys damage is much higher and the overall feel having two out is the same as before.
- Now receives second golem at level 10 (was 20)

### Iron Golem
- Rescaled to use damage curve, self scaling locked at 12+
- Added splash property
- Gains 1% PDR per level capping at 50%
    >This one is super squishable for some reason so hopefully this offsets that some. Will revisit if it doesn't.

### Fire Golem
- Fixed self damage scaling
- Now receives second golem at level 10 (was 20)

### Revive
- Now uses damage curve and uses the correct Skeleton Mastery field for synergy scaling

## Sorcerer
- Sorc orbs can now be used with Poison Strike, Sacrifice, Vengeance, Bash, and Carnage to accommodate oskill items


Bakelt:
excel/base/monstats.txt appears to have no changes while exce/monstats.txt did change

Delegus:
Yes this is still in draft.

Delegus:
Flipping to green, test before merge, don't yeet it. They feel underpowered to me due to slow attack speed or something, not sure yet. But the concept is done, balance just needs tuning maybe.

Bakelt:
Time to level a necro to test this weekend. Will give me something to do while I battle trying to find ideas for boss skills.

yinyin333333:
I went back to meet the spiders and used the skills with the same save file I used when I reported the Barbarian issue previously, but the frame drops I experienced then don't seem to be happening anymore i guess.

Delegus:
> I went back to meet the spiders and used the skills with the same save file I used when I reported the Barbarian issue previously, but the frame drops I experienced then don't seem to be happening anymore i guess.

Yeah that's where I went, but I tripled the density first. That does lag a little but no longer down to 5 FPS. 🤣

Bakelt:
Updated for playtest. Overall, it feels fine now once golems get up and running. May be a tad on the weak side due to golem attack speed but generally feels playable.

Bakelt (approved):
LGTM

PR #1233, opened 2026-05-08
no no no stack icon shadow

yinyin333333:
assassin has charge stack icons.
And Shadow Warrior and Shadow Master have the same stack icons.

gif1

Yeah... the icons are just floating all over the place.

And now only the Assassin will have the stack icons.
Advantages and disadvantages of making this change.

Disadvantages
- Other overlays applied to Shadow Warrior and Shadow Master, including the stack icons, are also not visible. This could be a disadvantage for those who frequently monitor the overlays of the Assassin's summons.

Advantages
- Everything besides the disadvantages. Those icons used to make me avoid summoning altogether when playing the Assassin, but now I'll be able to enjoy playing with summons normally.


The-Honorable (approved):
lgtm

PR #1239, opened 2026-05-12
Valk commando

Cantrip-commits-crimes:
Now using GenericPetAI!

Enabled: 
Able to use Skills (Normal Attack atm, but able to add later)
Run Animation (uses normal Amazon run animation)
Good AI (engages with a run, controllable SHORT tether)


yinyin333333:
web
transitions between animations look more natural and better as well. good

The-Honorable (approved):

PR #1242, opened 2026-05-13
fix bk tribal guardian visual

The-Honorable:
swap set visual for bul kathos tribal guardian to flamberge instead of war sword as the base changed in a previous patch to a colossus sword from mythic sword

image


Bakelt (approved):
LTGM

PR #1243, opened 2026-05-14
assassin trap rework/balances

The-Honorable:
- adjusted the spawning collision of all the sentries so that they can properly be placed under enemies

ice barrage
- increased cold duration from 2s to 3s to account for the hell cold length divisor

snowlash
- very much over performing, lowered damage curve in a few spots
- slightly increased the missile velocity
- slightly lowered the initial missile range
- slightly lowered the active range on the trap

glacial vortex renamed glacial burst
- reworked to release burst of glacial spikes in an area around the trap (similar to how fend projectiles work)
- has low hit delay, which makes it very strong for single target damage

fire blast
- slightly lowered the damage curve 

wake of fire
- removed wall collision on the spawner so small rocks dont interrupt the wave anymore
  - should greatly improve the clear in certain areas 

charged bolt sentry
- slightly lowered damage curve

lightning sentry
- slightly lowered damage curve

death sentry
- size from 1 to 3 so it will hit things in a bigger radius around the missile
- adjusted velocity/duration so it's slightly bigger aoe/pulse


Bakelt (approved):
LGTM

PR #1246, opened 2026-05-15
Colossal Nerfs

Cantrip-commits-crimes:
Colossal Jewels lose all CTC, were either broken (not working), harmful (Enchant/Hex for some builds), or too OP (BO).


yinyin333333:
image
Many who love CTC will shed tears like this. lol

The-Honorable:
> image Many who love CTC will shed tears like this. lol

I think the only one that was seeing ALOT of use was the Battle Orders one. The others seemed more Niche or did not work at all apparently. but i personally like CTC, but as they are currently designed i can see the desire to change them.

yinyin333333:
image

Well, looking at it this way, it does seem a bit ambiguous.
complaints I often heard from others about that Unique Jewel were that it is annoying because it overwrites the skill when the Unique Jewel's CtC level is lower than the Barbarian's skill level. I am not sure about other jewels.

The-Honorable:
> image
> 
> Well, looking at it this way, it does seem a bit ambiguous. complaints I often heard from others about that Unique Jewel were that it is annoying because it overwrites the skill when the Unique Jewel's CtC level is lower than the Barbarian's skill level. I am not sure about other jewels.

yea they ended up being niche and not great for the builds that had their own Skill and the CtC caused issues. the BO was good on a merc for non barbarians though.

Delegus:
Overall net nerf as is, would like to see something useful added to each one or the PR closed if you don't feel like doing it yourself.

Cantrip-commits-crimes:
It's not really a nerf, it's making them usable/good for more builds. As is, a lot of builds becomes WORSE from the triggered skill atm (aside from the phys one).

Cantrip-commits-crimes:
I legit had to disable the trigger to use them on multiple builds or they'd kill my build (disable Hex, etc).

Delegus:
Put the work in the make them interesting. 🤣

Delegus:
Closing this as we implemented something similar in 3.0.12

PR #1249, opened 2026-05-20
minigame

yinyin333333:
Draft

maps used in the Lab all have an indoor feel, consist of many floors, and are endgame content. They could be used to fill the gap between Pandemonium Tristram and the Lab, or later for mapping and reworking side areas (like the Mausoleum); in any case, if I just keep creating them whenever an idea comes to mind, they will definitely be useful somewhere down the line.


yinyin333333:
image
webp

Thinking of Maze Runner, I thought it would be nice to have a basic map of this type.
The corridors are very wide, which will allow for satisfying use of area-of-effect skills, and the size of the map itself is about 5 times wider and 5 times longer than the Tristram map, which is about 20 to 25 times the area.
advantage is that when Terror Zones are active, the monster density provided by the map's size would be enough to make running just this single map worthwhile, or it could be utilized for other custom level elements.

I increased the size of each map because maps that are too short end too quickly.
However, I didn't want to just stack many floors, as the Lab already uses that structure; I wanted to differentiate the feel by increasing the size of each individual map.
I’m also considering creating about 4 to 5 map concepts, with two short variations each for this large-scale map type—similar to the Azkaban map I created previously—so that they can be rotated and used later as mapping elements.

Of course, this is just my think and a draft, so it can be discarded at any time.

yinyin333333:
The town maps are currently just drafts where I've only established the connections, and I still need to clean up and remove the walls that cause unnecessary collision detection. The Act 2 and Act 3 maps are unstable areas that crash if you go outside a certain boundary, so I've temporarily blocked off parts of them.

Although they are familiar maps, ironically, monsters don't spawn in town, so I’m thinking I might be able to put them to some use in my own way.

PR #1250, opened 2026-05-20
TZ json compatibility update for D2R 3.2

yinyin333333:
act-specific presets for Terror Zones will be kept the same as in Reimagined.


Bakelt:
Changes look reasonable but would prefer someone with DLC to test and confirm how Herald is doing

yinyin333333:
> Changes look reasonable but would prefer someone with DLC to test and confirm how Herald is doing

changed values were taken directly from Retail.
Currently, the 3.0.9 JSON files are missing zone_data_id and weight. In Rotw 3.2, the only levels missing these values are areas like the Forgotten Tower—places with no monsters.
Since Herald accumulation is tied to hunting monsters in levels linked to those IDs, the absence of those IDs suggests that level progression itself might not be increasing. Some Korean users have also mentioned that Heralds do not appear in Terror Zones.
Cantrip said he was going to test it either yesterday or the day before, but he has disappeared.

yinyin333333:
https://github.com/user-attachments/assets/46b68d69-3af2-44e9-9fdc-c0e56238f863


When i test increase spwan rate for fast test, it json work.

The-Honorable (approved):

PR #1263, opened 2026-05-28
fix set bonus typo

The-Honorable:
should not be variable/range

aldurs full set now gives +2 max grizzly bears


Delegus (approved):
Needs most bearz most powah moar

PR #1274, opened 2026-06-01
Raven's Set level rebalance

strangemolars:
Reasoning:

Raven Set feels completely fine for endgame with the current power level of the mod. It equip level is too low (lvl 60 highest piece) and its power jumps are too staggered (lvl 45 in particular with 2 items and piles of resist and skills).

With Orbs being available, and full endgame sets being particularly easy to obtain during normal gameplay, I like the idea of balancing these in a more progressive manner regarding power.

So I'm tackling the Raven set thinking in how it would play in progression, same way I did with the Panda set.

Endstats of the set is basically the same (it has a fixed 9 raven hits instead of 8-9 and blind target 2 instead of 1), but it spreads its power-spikes a little bit more and make it so you can only wear the full thing at a more appropriate lvl 80.

Changes made:

Partial Set Bonuses:
40% all resist 3 piece set bonus split into 20/20 for 2-pc and 4-pc. 
+5 to Ravens moved to 3-pc

Full Set Bonuses:
20% all resist bonus added

Raven's Talons (gloves) (lvl 10)
30% cold resist 2-pc bonus split into 15/15 2-pc and 3-pc

Raven's Feet (boots) (lvl 10)
20% all res 3-pc removed (replaced by another equivalent partial set bonus and added to full set bonus) 
100 base defense split into 25 base and 75 as 3-pc bonus

Raven's Wing (shield) (lvl 45)
2 raven hits added
2 all skill 3-pc bonus removed
2% max cold resistance 4-pc removed (added to armor) 
30% cold resist added as 4-pc bonus

Raven's Beak (weapon) (lvl 60)
Item level from 35 to to 50
Equip level from 45 to 60
1 all skills to 2 all skills
1-2 raven hits to 2 raven hits
FCR from 20-30 to 10-20
1 Blinds Targets to 2
10% FCR 2-pc bonus to 20% FCR 3-pc bonus
2 raven hits 4-pc bonus removed (moved to shield)

Raven's Feathers (armor) (lvl 80)
Item level from 60 to 75
Equip level from 60 to 80
1 all skills added
30% cold resist removed (added to shield 4-pc bonus) 
3% max cold res 4-pc bonus to 5% max cold res

(Ben consulted. No birds were harmed during the making of this PR)


data/global/excel/base/itemstatcost.txt
```diff
@@ -389,7 +389,7 @@ stacked_gem	386		0	16								1	0	0	1	0		16	0	16	0																56	19	0	Stacked
 item_nonclassskill_display	387		1	15	9						1			1	181	327		7	20	6	0	9															150	28		ItemModifierNonClassSkill	ItemModifierNonClassSkill										0
 ravensetwolf	388		1	5										0		0		5	0	5	0																298	19		ravensetwolf	ravensetwolf										0
 ravenhits	389		1	5										0		0		5	0	5	0																298	19		ravenhits	ravenhits										0
-ravendmg	390		1	5										0		0		5	0	5	0																298	19		ravendmg	ravendmg										0
+ravendmg	390		1	16														16	50	16	50																298	19		ravendmg	ravendmg										0
```
Delegus: You can't make these kinds of changes, it breaks the character files when you mess with save bits, after a quick glance it looks like the rest of this PR is based on this change and so I'm not going through the rest of it yet, you're likely going to need to create a new stat or work within the bounds of this one.

strangemolars: nah, in that case the best course of action is lust not making the stat change now. if you guys ever want or need to make a variable %raven damage stat for some reason, then you do it with a new stat like you mentioned or at a major patch change so that breaking old saves with the affected items dont happen. I'll revert it to the old logic tomorrow. the rest of the PR is not dependent on it. when I was testing it I was always spawning new items on clean saves, so I never realized they would break saves with old versions of these items.

data/global/excel/sets.txt
```diff
@@ -103,7 +103,7 @@ Celestial Hierarchy	Celestial Hierarchy	100	mag%		20	20					mag%		25	25					mag%
 Lords of Hell	Lords of Hell	100	allskills		2	2																													pierce		75	75	gethit-skill	Charged Bolt	13	31	slow		20	20	res-all		40	40	noheal		1	1	openwounds		55	55										0
 Corruption Coils	Corruption Coils	100	regen		-50	-50																													res-all		-50	-50	addxp		-40	-40	extra-pois		40	40	mag%		100	100	allskills		5	5														0
 Maadi's Paradox	Maadi's Paradox	100	res-all		50	50					skill	Bone Spear	5	5					hp		200	200													allskills		5	5	nofreeze		1	1	oskill	Energy Shield	10	10	res-mag		15	15																		0
-The Raven's Nest	The Raven's Nest	100	skill	Raven	5	5	ravensetwolf		1	1	res-all		40	40					ravensetwolf		1	1													ravendmg		1	1	oskill	Amplify Damage	1	1	red-dmg%		20	20	abs-mag%		10	10	cast2		30	30														0
+The Raven's Nest	The Raven's Nest	100	res-all		20	20	ravensetwolf		1	1	Skill	Raven	5	5					res-all		20	20	ravensetwolf		1	1									ravendmg		1	1	oskill	Amplify Damage	1	1	red-dmg%		20	20	abs-mag%		10	10	cast2		30	30	res-all		20	20										0
```
Delegus: Skill needs to be lowercase, looks fine otherwise.

Delegus (approved):

PR #1279, opened 2026-06-05
Teeth buff + pokes on some items

strangemolars:
Reasoning:

Teeth 2 weak, needs to be stronk. 
Ran it a lot in the past week trying to get a point where it feels better but not clear cut just better than Spear. It has a way bigger spread, but since the projectiles dont pierce (I tried making Teeth pierce, felt too similar to Spear, so I reverted back) and the damage is still lower, I didn't notice bigger improvements in clearing speeds. But both skills feel viable now.

Poked some items to try to make the Spear items more competitive with Maadi. Still think Maadi is better overall for someone focusing Spear, but at least, theres something to consider.

While still trying to nudge people towards trying items for Bone necro other than Maadi. removed Spear bonuses and buffed the Teeth ones from Boneshade to kinda signal towards that on the itemization. Also made a Redemption shield as a possible alternative on the unique-gear profile for the build considering its interaction with the life-casting mechanics and whatnot.

Also rebalanced the Maadi levels following the logic on the other ones I did. This one had a quite nice spread from the start, just needed to be a bit higher level for full set :3

Changes made

Skill

Teeth
Velocity from 16 to 20
Starting projectiles +1 (starts at 3)
Sinergies reduced from 13% to 10%
Damage curves increased (avarage damage):
Base: same
2-8: from 1 to 2
9-16: from 3 to 5
17-22: from 6.5 to 10
23-28: from 11 to 15
29+: from 15 to 23
This results in a buff to damage curves of about 65%, which with the nerfs to synergies bring the spell to around 45% of Spear damage and an overall 40% damage buff.

Items

Pull of Darkness
2 all skills to 2 necro skills
3 random skill to 2 random skill
2 Bone Spear added

Boneshade
3-5 to Bone Spear to 1-3
3-5 to Teeth to 5-7 Teeth

Devil's Advocate
3 to random skill to 2 random skill
3 to Bone Spear added

Reaper's Trophy
2 all skills to 1 necro skills
2 poison and bone skills added
Slain Monsters Rest in Peace removed
20 FCR added
Lvl 3-4 Redemption Aura when equipped

Maadi's Set

Partial set bonuses:
50% all res 2-pc split into 25/25 2pc and 4-pc

Maadi's Vision (headslot)
Equip level from 63 to 65

Maadi's Torch (wand)
ilvl from 64 to 70
Equip lvl from 64 to 73

Maadi's Soul (shield)
ilvl from 49 to 75
Equip lvl from 49 to 81
Extra bone spear projectiles moved from 5-pc set bonus to base item

Please test new Teeth damage in particular. I think its absolutely fine, specially with the building consideration that exist within the bone tree, but I might have missed something.

EDIT: forgot to list the change to mana cost on Teeth. Made it be .5 per level as the other 2 bone skills instead of .75. It ended up costing significantly more than even Spirit by the end of the game. Also remade the Teeth and Spear numbers to shift7 instead of shift5.


Bakelt:
Probably won't get to testing myself but only concern I really see is the substantial buffs for 1-16. I feel like that will just be screen clearing the first two acts

strangemolars:
> Probably won't get to testing myself but only concern I really see is the substantial buffs for 1-16. I feel like that will just be screen clearing the first two acts

Oh, I mistyped on the summary that the avarage damage went from 1 to 3, but it went from 1 to 2 (curves were 1min 1max I changed to 1min 3max on the increases). 

It is still a big increase. I thought it was mostly on par with what we see nowadays in the mod, but if you guys think its too much I can recurve differently.

data/global/excel/base/skills.txt
```diff
@@ -66,7 +66,7 @@ Lightning Mastery	63	sor	lightning mastery
...
+Teeth	67	nec	teeth		8													teeth																																																																															necromancer_bone_cast														19	17						teeth	teeth	bonecast										1	1		none													SC	SC	xx																				1							necromancer_bone_cast			1	25														1	1				1	1	7	6	1	1	1											min(ln12,25)	# of Missiles created	par3	Missile Activation Frame	0	# of missiles that can trigger item modifier events															3	# of Missiles created baseline	1	# of Missiles created per level	0	Missile Activation Frame							2	Energy per 1% Synergy	10	Damage synergy																									1								8															mag	2	1	3	7	13	19	4	3	7	13	21	27	(stat('energy'.accr)/par7)+((skill('Bone Wall'.blvl)+skill('Bone Prison'.blvl)+skill('Bone Spirit'.blvl)+skill('Bone Spear'.blvl))*par8)								1	1	0
```
Delegus: IIRC 24 used to be set here because 25+ caused the multiple shot function to crash, did you test and make sure this was no longer the case? Also with that many projectiles whether it is even or odd is insignificant.

data/global/excel/base/uniqueitems.txt
```diff
@@ -1221,8 +1221,8 @@ Vow of Revenge	1215	100		1	1				1	1	35	47	ne9	cantor trophy		5	5000								ac%
...
 Venomlord's Visage	1218	100		1	1				1	1	43	55	neb	minion skull		5	5000	dgrn	dgrn						ac%		150	200	oskill	Inferno	15	20	res-fire		35	50	allskills		1	1	demon-heal		20	20	skill-rand	3	66	95	skill-rand	3	66	95	addxp		3	5	oskill_hide	Hidden Charm Passive	1	1														0
-Reaper's Trophy	1219	100		1	1				1	1	39	49	neb	Minion Skull		5	5000								ac%		130	180	block		20	30	block2		20	30	allskills		2	2	hp/lvl	12			mana/lvl	12			res-cold		45	45	rip		1	1	oskill_hide	Hidden Charm Passive	1	1														0
-Devil's Advocate	1220	100		1	1				1	1	70	75	neg	hellspawn skull		5	5000				invnec10				extra_bonespears		1	1	cast1		20	30	nec		1	2	extra-mag		15	20	ac		150	200	skill-rand	3	66	95	res-all		20	30	oskill_hide	Hidden Charm Passive	1	1																		0
+Reaper's Trophy	1219	100		1	1				1	1	39	49	neb	Minion Skull		5	5000								ac%		130	180	block		20	30	block2		20	30	nec		1	1	hp/lvl	12			mana/lvl	12			res-cold		45	45	skilltab	7	2	2	cast1		20	30	aura	Redemption	3	4	oskill_hide	Hidden Charm Passive	1	1						0
```
Delegus, line 1224: Redemption on a necro usually feels bad, regardless of spec, but just pointing it out as a consideration, I am indifferent to it either way.

data/global/excel/base/skills.txt
strangemolars: > IIRC 24 used to be set here because 25+ caused the multiple shot function to crash, did you test and make sure this was no longer the case? Also with that many projectiles whether it is even or odd is insignificant.

oh, just saw this one now. Yeah, I did level the char with it and did a lot of HTZ cow runs to compare it with Spear in terms of clearing, and I didnt get any crash.
but you are right regarding the numbers, I can just leave it at 24 just to be safe if this is a problem that happened in the past.

data/global/excel/base/uniqueitems.txt
strangemolars: > Redemption on a necro usually feels bad, regardless of spec, but just pointing it out as a consideration, I am indifferent to it either way.

Yeah. The original item already had RIP, which already had that same problem but without any upside. I thought of making it Redemption so it feels like a choice for bone as in that it cockblocks you a little bit in terms of using Revives and Corpse explosion (although you can still do if you are quick) but it gives you more sustain for the Bloodcast mechanic.

data/global/excel/base/skills.txt
strangemolars: Reverted to max 24 just in case. Kept starting at 3. Changed desc field to reference the calc field and tested.

data/global/excel/uniqueitems.txt
```diff
@@ -351,7 +351,7 @@ Dragonscale	347	100		1					1	1	84	80	pae	zakarum shield		3	5000	dgrn	dgrn						a
 Steel Carapace	348	100		1					2	1	60	66	uul	shadow plate		3	5000	dgry	dgry						ac%		190	220	balance2		30	40	ease		-25	-25	red-dmg%		10	15	red-dmg		10	15	gethit-skill	Inner Sight	10	12	aura	Thorns	8	10																						0
 Medusa's Gaze	349	100		1					1	1	84	76	uow	aegis		3	5000	lred	lred		invtowu				ac%		150	180	hit-skill	Decrepify	9	5	gethit-skill	Decrepify	15	5	lifesteal		5	9	res-all		30	50	oskill_hide	Hidden Charm Passive	1	1																										0
 Ravenlore	350	100		1					1	1	82	74	dre	sky spirit		3	5000	dgld	dgld						ac%		120	150	res-all		15	25	skilltab	17	3	3	enr		20	30	pierce-fire		10	20	skill	Raven	7	7																										0
-Boneshade	351	100		1					1	1	85	83	7bw	lich wand		3	5000	dgry	dgry		invbwnu				nec		2	3	cast2		25	35	skilltab	7	1	2	skill	Teeth	3	5	skill	Bone Armor	3	5	skill	Bone Spear	3	5	skill	Bone Spirit	3	5	extra-mag		10	15	oskill_hide	Hidden Charm Passive	1	1														0
+Boneshade	351	100		1					1	1	85	83	7bw	lich wand		3	5000	dgry	dgry		invbwnu				nec		2	3	cast2		25	35	skilltab	7	1	2	skill	Teeth	5	7	skill	Bone Armor	3	5	skill	skill	Bone Spirit	3	5	extra-mag		10	15	oskill_hide	Hidden Charm Passive	1	1																	0
```
Delegus: The properties get out of whack here, they are too far to the left starting with 6/7 ish area. 

Image

Delegus:
Conflict fun

Delegus (approved):

PR #1281, opened 2026-06-06
begone low quality

The-Honorable:
removes low quality from dropping. seems to work. based on an old pk post.

https://d2mods.info/forum/viewtopic.php?f=4&t=40223


yinyin333333:
I have already been applying this change to the KR files for some time, and it has been confirmed that only the low quality was removed without any other issues.

The-Honorable:
> I have already been applying this change to the KR files for some time, and it has been confirmed that only the low quality was removed without any other issues.

did you apply any other item ratio changes?

Delegus:
Hater

yinyin333333:
no just  for low(crack)

yinyin333333:
I did not know the translation was weird because I was playing online with other user.
I only modified norm, and I did not modify anything else.

Delegus (approved):
Hating on the less fortunate

PR #1285, opened 2026-06-10
terrorzone weather,day,night fix

yinyin333333:
About two months ago, I submitted changes regarding the purple tint and the darkness in Terror Zones. #1161 

That solution involved forcing a specific weather state (daytime) onto the Terror Zones. While that was certainly a valid solution, I didn't think much of it at the time. However, while helping someone who uses a different mod, they asked if it was possible to make the Terror Zone weather dynamic instead of fixed, and that got me thinking.

And I fixed that, so now it doesn't force a single weather type, and I've modified it so that TZs no longer interfere with the lighting and weather. Naturally, the purple tint is removed, and the natural transitions of time and weather while playing are preserved.

I was at work, so I couldn't test it for a long time, but after leaving it on in windowed mode for about 15 minutes, I confirmed that the Terror Zone successfully transitioned from a bright daytime screen to a dark nighttime one.

**_While this is the most suitable solution for the original purpose, the darkening-related options in the launcher will also need to be adjusted._**


Delegus:
I see, as it is the launcher would likely still work fine, it just tries to remove those if they exist, lolz. Can update it though.

yinyin333333:
Ah, that makes sense. Those files didn't exist before in reimagined, after all.

However, since the vis_event_map.json file in the data\hd\env\weather directory was modified rather than deleted, it might need to be updated. (Or does the launcher overwrite this with the original from a clean folder too? I'm not sure about that. Always grateful for your work. Keep it up.)

Delegus (approved):
Works as advertised is appears

PR #1286, opened 2026-06-17
MoltenBoulder-FirePathBugfix

Bakelt:
- Fixed .bvl to .blvl (ty shooby)
- Copied the fire DoT curve from armageddon


The-Honorable (approved):
lgtm

PR #1288, opened 2026-06-23
Herald spawn linear

yinyin333333:
current PR is set to 1%.
Looking only at the values, the rates in retail (currently Reimagined 3.0.10) may seem higher than the 1% in this PR. However, after the 3.2 update, the spawn rate was changed to increase in a curve based on level progression, and while it varies by tier, it is effectively 0% for the first 50% of progression in most tiers.
This PR, on the other hand, uses a flat, linear structure that is fixed at 1% regardless.

It is not strictly necessary to change it.
The main issue is that after the Rotw 3.2 update, both retail and Reimagined users feel that while Heralds do appear, they seem somewhat similar to before (when level progression is high enough), but also feel a bit lower (when level progression is insufficient). Therefore, it is best to try to find that original feeling where the spawn rate in Reimagined was slightly higher than in retail.

Please test it and let me know.


yinyin333333:
When i test
It definitely feels like changing it according to the PR brings out the right feel a bit better, but I'm not really sure if sticking with the existing way feels particularly lacking.
And since the number of complaints regarding the Herald drop rates after version 3.2 is extremely small, this PR doesn't seem to hold much significance, so I'll close it for now.

Delegus:
Is it being open costing us something? 🤣

yinyin333333:
> Is it being open costing us something? 🤣

Oh, it's not that, it's just that looking at open PRs triggers my OCD a bit.

Delegus:
Lmao, yeah me too, so I just don't get on the page until I'm ready to review them. 🤣

PR #1289, opened 2026-06-23
Arrow# on some skills made softpoints

strangemolars:
Reasoning:
Been testing a few build the last couple months and making cold/fire/magic and freezing/exploding arrow# scaling as softpoints enables some cool archer variants for other classes while not impacting amazon leveling in a significant manner.

These changes enabled things like ranged enchant sorcs and ranged holy aura paladins to have other competitive options for their builds instead of basically only using Multishot on non-synergistic bows/xbows. None of the builds were even close to being problematic and supplanting the place of the original skill wielder, but they were functional and a cool thing to play.

Elemental bow amazon class identity doesnt feel too much at risk with both of them capitalizing on the higher level arrows in Glacial Orb and Meteoric arrow and not the lower level more "generic feeling" arrows. (I feel Glorb and Meteoric should be kept as much as possible restricted to Amazon, so 'freeing' the lower level arrows to other classes shouldnt be a problem. There are Glorb and Meteoric oSkill items present, but I don't think they are good enough to enable builds for other classes).

Changes made:

Magic Arrow, Cold Arrow and Fire Arrow:
- All made 1 arrow at base level, then 1 extra arrow at level 3 softpoints and 1 extra 3 after that, capping at 9 arrows max (same as when it was hardpoint self synergy)
- Made the 3 central arrows leech and trigger effects, up from the middle 2.
- Currently fire/cold arrow dont matter much as there are no oSkill with them, but I plan on adressing this in the future.

Exploding/Freezing Arrows:
- Both made 1 arrow at base level, then 1 extra arrow at level 10 softpoints and 1 extra at level 20 softpoints after that, capping at 3 arrows max (same as when it was hardpoint self synergy)
- Made only the central arrow leech and trigger effects, down from the middle 2. Since these two arrow skills carry elemental fire/cold damage on its explosions and the items that have it have also other good secondary sources of damage (as in ctc proc on harp of the infernos and holy freeze aura on chillwind), I felt they didnt need to trigger more stuff to still have decent damage.

OBS: I have files for a PR that will tackle some reworks/adds for quivers and some bowx/xbows I've been working on with Ben's supervision to fill some itemization gaps these kind of chars, but I prefer this one was a standalone PR since its just beneficial even without anything else being fondled with.

There is still one build-enabler item for Amazon I want to test more deeply before submiting all these items that I'm working on.


Delegus (approved):
Don't think this will affect Amazon too much aside from making hardpoints synergy engines but that's all they are anyways at the moment, may create an issue to give the skills more individuality later, we'll see.

PR #1292, opened 2026-07-08
maggot floor janitor

yinyin333333:
image

The floor was a bit messy in the previous Maggot Lair expansion.
It doesn't significantly affect gameplay and the fact that it expanded is what matters, but I have made the floor cleaner and more normal. There might be some parts I missed... well... I think it's better than before.


Delegus (approved):
This looks excellent in game.

PR #1298, opened 2026-07-28
player dummy

yinyin333333:
Hmmmmm I'm not sure how it'll be used (or it might not even be used at all), but I added it as a dummy since I think Ben mentioned it before.
The monster AI is just a placeholder too, and since the skills are currently using the existing player skills, they'll obviously need to be changed. (and meesy setting)
Don't just merge it blindly—take a look and use it if you want, Ben. If not, you can just close it. I'll leave it up to your judgment.


yinyin333333:
its not draft. Ben said he's going to use this somewhere else.

The-Honorable:
thanks they will be wandering in lab. ill give them unique skills and such in my branch

PR #1299, opened 2026-08-04
disable lightbeam and few extra string

yinyin333333:
Yes, I understand. Some people might like this change, while others might not.
If you think this change is too much, feel free to close it.

But honestly, there are way too many light beams. I'll leave only some of the higher-tier runes and remove the rest.
And in my personal opinion, having "pick up" in the text string doesn't look clean and feels like it's telling me what to do.
I'm only going to pick up what I want to pick up. (Though I know no one was actually forcing me. Just kidding)

- Disabled drop-light visual effects for selected runes, keys, quest items, power orbs, storage items, grabber tools, and jewels.
- Removed localized pickup prompts such as ~Pick Up~ things from rune strings.
- Runes from Gul Rune onwards will keep their light beams. I could remove them all, but seeing something like a Ber Rune dropped is always a pleasure—whether in legacy, Resurrected, or Reimagined. It also fits the whole purpose of having a light beam, showing that something good has dropped.


yinyin333333:
image
See you in the launcher options. bye

PR #1302, opened 2026-08-09
Sigil swap

### LINKED ISSUE #1240: Sigil:Death / Sigil:Lethargy animation "bug"
Sigil:Rancor uses the Fire visuals (which are what Sigil:Death should be doing)

Image

while Sigil:Death is Red (like what Sigil:Rancor is doing)

Image

--

The symbols are matched with the skill icon, so the play is swapping the skill icons between the 2 (and recolor), and then swap the seals.

yinyin333333:
close #1240 

These are the changes Cantrip wanted.
swapped the visuals, the casting overlays, the icons, and the icon colors.


Delegus (approved):
Image
Works as advertised, nice work.

PR #1306, opened 2026-08-19
Blizzard 3.3 compatible

yinyin333333:
states.txt

* Updated the existing states to match the hardcoded row index of the newly added `primeevil_threat`.

skills.txt

* Warlock's Sigil Death `etype fire` has already been applied, and some values for `bind demon` were changed.

propertygroups.txt

* Honestly, I don't think this is necessary. It probably won't be used even if it's included, but I added one row anyway to avoid confusion in the next version or the version after that.

remaining .txt changes were excluded because they were unnecessary vanilla-related changes, such as drop TCs.


data/global/excel/states.txt
```diff
@@ -230,60 +230,61 @@ bind_demon_underling	227																																																		19
 herald	228																																										herald																															0
 apocalypse	229																																																																									0
 missilelimit	230																																																																									0
-elementalcovenant	231					1																	1	1	1	1																	elementalcovenant							lightresist																								0
```
Delegus, line 233: Your going to want to just move elemental covenant to the bottom since it is a skill only state, so you aren't pushing around other states for other items because it's going to bork them.

yinyin333333:
> Your going to want to just move elemental covenant to the bottom since it is a skill only state, so you aren't pushing around other states for other items because it's going to bork them.

I changed it as you suggested. As far as I can tell, nothing currently references it by row index, so leaving it where it was probably wouldn’t cause any issues, but there’s no particular reason not to move it either.

Delegus (approved):

PR #1307, opened 2026-08-19
Druid charge system + other things

strangemolars:
---BEHOLD THE TEXT WALL AND DESPAIR---

WOLF FORM: 
Added a 0-70 (10% lvl 1, 61% lvl 25) velocity curve to baseline WW form (removed from Feral Rage)

Charge mechanic
Current: 3 starting charges, 1 additional charge at lvl 2 and 2 levels after that. Softpoints included, capping at 53 charges.
New: 2 starting charges, 1 additional charge at lvl 4 and 1 every 3 after that. Hardpoints only, capping at 10 charges. Maybe items can be added that increase max charges (thoughts).

MAUL
Attack rating from 40% starting 15% per lvl to 50% starting 15% per level (no reason for it to be this weird 40% one only when shockwave, fireclaws, Arctic claws, etc arre all 50/15)
Single Target stun 0.4-4s diminishing returns curve kept as is.
Current: 3% eIAS and 30% ED per charge, max 53 charges.
New: 1% Crushing Blow per charge
3% eIAS base +1% per 10 levels per charge
20% ED base +1% per level per charge

-Relative full charged progressions for comparison (with 25 hardpoints)
LVL 1
Old Maul: 9eIAS/90ED
New Maul: 6eIAS/42ED/2CB
LVL 5
Old Maul: 15eIAS/150ED
New Maul: 9eIAS/75ED/3CB
LVL 10
Old Maul: 24eIAS/240ED
New Maul: 20eIAS/150ED/5CB
LVL20
Old Maul: 39eIAS/390ED
New Maul: 40eIAS/320ED/8CB
LVL 25
Old Maul: 45eIAS/450ED
New Maul: 50eIAS/450ED/10CB
lvl 35
Old Maul: 60eIAS/600ED
New Maul: 60eIAS/550ED/10CB
LVL 50 (This is around the lvl I see most physical chars sitting a if they are not stacking Heaven’ Facets, which they shouldnt Imo).
Old Maul: 84eIAS/840ED
New Maul: 80eIAS/700ED/10CB
LVL 75
Old Maul: 120eIAS/1200ED
New Maul: 100eIAS/950ED/10CB
LVL 100
Old Maul: 159eIAS/1590ED
New Maul: 130eIAS/1200ED/10CB

FERAL RAGE
Current: 10%-70% FRW diminishing returns curve and 4% life leech per charge, max 53 charges.
New: FRW scrapped. It worked as I intended with new charge curves and all, but honestly feels like a dumb stat to have on this skill. Prefer tweaking the forms base speeds.
2% Life Leech per charge - this basically means it goes from a perma 100%+ leech for wolf down to “only” 20% (maybe a bit more If I make +charge items), which is a big nerf, but I think its in line with current filosophy regarding Leech. It might also even feel too much to what you guys aim, but considering all the difficulty modifiers and stuff, it ‘should’ be acceptable?
2% -Enemy Physical Resist per charge (Deadly Strike honestly sounds more flavorful, but also not a strong enough thing on a build that needs some help, and something you can get to very high values already so it would be a very low power increase unless is big numbers).
-2 Enemy defense per hit +1 every 2 levels per charge – on odd levels (sendbits changed to make it go over -128, tested and seem to be working, at least on the char advanced stats. I choose the one because its a different way to make up for AR considerations that feel it fits with the build flavor).
10% ED + 1% every 2 levels per charge – on even levels. I tried a lot to think of something different, but barring remaking Feral Rage its own attack skill in a way and scrape the entire synergistic thing it has going on the build, I don't see another good stat to scale with levels here. I'm trying to make it not so silimar to Maul by having it go "hafsies" in here with -enemy defense per hit and all, but I get it it would be cooler if it was something else. 
Base skill damage scrapped. Its really meh and weird to be here, and also not needed after the aurastate ED% thing.

Considerations looking at Fury physical builds:
Anyway you slice it, ‘forcing’ players to spend points into Feral Rage and making it a more 'official' part of a Fury build comes with some costs. Its fixating 25 points on the build that before could go anywhere as the one (and only IMO) benefit Fury had over more regular physical builds is that it was very light on points. You really only NEEDED the 25 points in Fury, and everything else is up to you.

I think most people agreed that Lycan was kind of a must. After that, most would also put into Barbs, although here theres more variation because some people have a hard time realizing how good -enemy res even after so many years. I dont see those changing since they should still be the optimal choices to make. 

That left 25 free points. Most people I think spend some points into Bears and Oak, or the ones prone to fall into traps would put into the actual Werewolf form (something that was not good to do even when it gave 15%AR per level, and now that it gives 3% only, less than HALF of the AR that HotW gives per level, and virtually no other benefit since IAS is on diminishing returns curve while HotW also gives ED and eIAS. So people making this choice completely baffles me).

(Addendum: Feral Rage being a thing will open the option to a 'glass cannon' wolf with no Lycan points and all out aggressive skills - Fury+Barbs+HotW+Feral Rage - if that would be better than letting go of one offensive skill for Lycan, I honestly doubt, but a choice is a choice, and someone is bound to make it someday).

I tried a bunch of stuff around and I set for believing that maxing Heart of the Wild was the by far the best option for those 25 points - although I never found anyone that agreed with me on this - not that there actually are many Fury players going around on this mod. (Ok, 3. My sample size was only 3 god damn it ROFL.) Ben in particular thinks the skill is quite bad (xD) but for me the reason it is actually quite nice is that its impacts in player's power is not bound by the 25 hardpoint synergy cap other physical skill synergies are limited to, so even if it starts low on its benefits, by the end it shakes up to be more than a usual physical skill synergy gives.

So to remediate that potential loss of damage, at first - I just thought of making Feral Rage a synergy for Fury and tweaking the numbers. But then I figured making it have ED% per charge like Maul does is a better way to make it an "indirect synergy" and more scalable with levels - and make more sense within Shapeshift mechanics. 

I know I'm kinda prolyx most of the time because I suck at writing (specially in english), but I hope it helps to explain why I felt swapping Feral Rage to a hardpoint system also implicitly implied in making it boost Fury somehow in return, since you would be losing damage in other fronts, and why I choose to go with an soft-point scaling mechanic akin to Maul instead of just making it a flat Fury synergy as it would be easier to do.

Below is how stuff match up when you consider Old Feral Rage,New Feral Rage and HOTW against each other. 

(Addendum: Consider that if a player decide to not put points into Feral Rage, it can still theoretically benefit from 2 Feral Rage charges that will scale -enemy def and ED% with soft point levels. Lets say a players gets 25 soft points – high number for these builds – that would mean at 2 feral rage charges 45ED, -28Enemy Def per hit, 2%-EPDR, 4%LL; while in the reversed scenario a lvl 25 softpoint HotW would give 135ED, 193AR and 5eIAS – assuming everything is working as it should).

-Relative full charged progressions for comparison (with 25 hardpoints – Enemy defense debuff might be a couple points up or down because I calc’d this before I made them be at odd level numbers so every point on feral rage grants some benefit instead of only even levels)
LVL 1
Old Feral Rage: 12 LL, whatever FRW
New Feral Rage: 20ED, -4EDH, 4%-EPDR, 4%LL
HotW: 15ED, 25AR, 6eIAS
LVL 5
Old Feral Rage: 16LL, whatever FRW
New Feral Rage: 36ED, -12EDH, 6%-EPDR, 6%LL
HotW: 35ED, 53AR, 10eIAS
LVL 10
Old Feral Rage: 24LL, whatever FRW
New Feral Rage: 75ED, -35EDH, 10%-EPDR, 10%LL
HotW: 60ED, 88AR, 15eIAS
LVL20
Old Feral Rage: 52LL, whatever FRW
New Feral Rage: 160ED, -96EDH, 16%-EPDR, 16%LL
HotW: 110ED, 158AR, 25eIAS
LVL 25
Old Feral Rage: 60LL, whatever FRW
New Feral Rage: 220ED, -140EDH, 20%-EPDR, 20%LL
HotW: 135ED, 193AR, 30eIAS 
lvl 35
Old Feral Rage: 80LL, whatever FRW
New Feral Rage: 275ED, -195EDH, 20%-EPDR, 20%LL
HotW: 185ED, 263AR, 30eIAS	 
LVL 50 (This is around the lvl I see most physical chars sitting at). 
Old Feral Rage: 142LL (LOL), whatever FRW
New Feral Rage: 350ED, -270EDH, 20%-EPDR, 20%LL 
HotW: 260ED, 368AR, 30eIAS
LVL 75
Old Feral Rage: 160LL, whatever FRW
New Feral Rage: 475ED, -390EDH, 20%-EPDR, 20%LL
HotW: 385ED, 543AR, 30eIAS 
LVL 100
Old Feral Rage: 212LL, whatever FRW
New Feral Rage: 600ED, -520EDH, 20%-EPDR, 20%LL
HotW: 510ED, 718AR, 30eIAS 

Where to go forward?
+charges items can be added to the game, mostly on druid pelts I’d say. They could be specific to one form or for both. As it sits, at skill level 50, 1 aditional charge would be: 35ED, -14 enemy defense per hit, -2pdr, 2ll for Wolf; and 60ED, 8eIAS and 1CB for Bears.

POISON DRUID

Ben asked for pokes on Toxic Fang. Went through a couple interactions, what we set for that achieves the intended goal of Fangs being the skill you spam on though enemies. Also did a couple rebalances on Rabies due to its playstyle.

RABIES 
Attack rating to 20% base and 10% per level (from 50/10). Just to make it a tid bit more unreliable and normalize AR numbers as well. Also, make it a bit worse on bosses so theres less incentive to use there.
Mana cost from flat 10 to 5 +1 per level. Considering by the end you bite once and run for 20 secs basically, you have time to mana regen. And MPK works just fine on the skill. If its not a spam skill, mana cost needs to be high imo.

TOXIC FANGS 
Toxic Fangs also carries the -20 enemy poison resist aura state Rabies has. It makes sense for me, specially on some weird luls where it felt the loss of the debuff makes very high resistant mobs take less damage from Fangs then from Rabies before you start to amass the stat.
Changed to a 0.4 seconds duration from 2 secs. Initial iterations went the other way, increasing it to 4 to allow for more weaving of other skills in the duration, but if the intention is spamming it on bosses, reducing made more sense.
Per-bit base damage basically increased by 300%. Since its just a straight poison spell and doesnt have any additional shenanigans like Venom does, it damage had to be cranked up significantly - and its still nothing to write home about on bosses compared to basically anything.
Synergies to 13% from 10%. Consistency and parcel of the damage buff needed.
Mana cost curve kept, but starting mana cost is 2 instead of .2, so its basically the same curve as Fire Claws.
AR to 15% per level, from 10%. Help on bosses in particular.
Added a flat physical damage curve about as double as current Feral Rage / Maul ones. With the increased mana cost, feral rage leech nerf, and since you gonna sit and spam the spell now - I felt this would help a bit with sustain a little bit. The skill has no ED% and tends to use very low damage weapons too.

Where to go forward?
Scrap Poison Creeper and make a Quillbear summon that has a Toxic Aura or some way to debuff enemy's poison res (move that aspect of the build to the summon instead of having it baseline on the spells); or some other way to be synergistic with be build beyond just damage synergy.

OTHER STUFF
Poked some other druid tooltips for consistency and clarity (particularly the placement of the mana-costs that were all over the place).


data/global/excel/base/skilldesc.txt
```diff
@@ -161,30 +161,30 @@ scroll of townportal	0	0	0	4	18		skillname219	skillld219	skillld219	skillan219
...
 shape shifting	2	1	2	2	52		Skillname225	Skillld225	Skillld225	Skillan225																	36	StrSkillDurationPlus		ln12	25																										74	StrSkill100		pst1																																																											0
 firestorm	3	1	1	3	32		Skillname226	Skillld226	Skillld226	Skillan226	8	2															75	StrSkillFireDamageAveragePerSecond		enms*25/256	exms*25/256																										36	StrSkill3		usmc	256																					40	Sksyn	Skillname226	2		76	Firedplev	StrSkillEle	par8		76	Firedplev	Skillname238	par5																								0
-oak sage	1	2	1	1	16		Skillname227	Skillld227	Skillld227	Skillan227																	36	StrSkill3		usmc	256	13	StrSkill42		clc1		36	StrSkillRadiusSingular	StrSkill18	ln78*2	3	74	StrSkill100		dm12																																																																										0
```
Delegus, line 167: Things on the left side are changing values per soft point, such as mana cost in this case, and on the right side are things that do not change or require a visual update when there are points to be spent. Moving it to the right will put the mana cost on top, but also remove the visual indication that mana cost increases per level. I may have missed some but I have made many passes over tooltips in order to ensure this rule is followed whenever possible. This change is unfortunately counterproductive unless you remove the scaling mana cost from the skills you moved it around on.

data/global/excel/base/skills.txt
```diff
@@ -231,17 +231,17 @@ Wearbear	228	dru	wearbear		116																			bear				damagepercent	par1 + bl
...
+Hunger	242	dru	hunger	16	32										hunger_splash										rabies			poisonresist	-par2																																																																																								hunger_splash													1	0		h2h													S3	S3	S3				1		1	1								1		1			1										30	25								2	wolf	bear				1	1				1	1	6	8	1	1	1											par1	%Dmg Phys Increased					par4	%dmg converted to elemental type													0	%Dmg Phys Increased	20	Pois Res Reduction			0	%dmg converted to elemental type							13	Damage synergy																									1	50	15						5	128	16	16	32	64	96	192	32	16	32	64	96	192		pois	20	6	10	12	24	32	30	8	14	18	32	42	(skill('Rabies'.blvl)+skill('Plague Poppy'.blvl))*par8	10							512	15360	0
```
Delegus: This is going to directly conflict with Rabies and may cause the Rabies missile to expire prematurely due to state clashing, is this intended?

data/global/excel/base/itemstatcost.txt
```diff
@@ -119,7 +119,7 @@ item_fractionaltargetac	116		1	8											67	20		7	0	7	0											1					118	19
 item_preventheal	117		1	8											48	50		7	0	7	0											1					81	19		ModStr4a	ModStr4a									2	0
 item_halffreezeduration	118		1	2											5096	988		1	0	1	0																19	19		ModStr4b	ModStr4b									2	0
 item_tohit_percent	119		1	12											981	40		9	20	9	20											1					117	19		ModStr4c	ModStr4c										0
-item_damagetargetac	120		1	8											24	-20		7	128	7	128											1					75	19		ModStr4d	ModStr4d									2	0
```
Delegus, line 122: Changes in sendbits are normally fine but from 8 to 12 is from 255 to 4095 and may be excessive. It's 2 ^ (# entered) - 1

data/local/lng/strings/skills.json
```diff
@@ -30296,19 +30296,19 @@
...
-        "frFR": "en forme de kodiak,\nmutile vos ennemis\npour augmenter les dégâts\nau fil des coups",
-        "itIT": "Quando sei in forma di orso mannaro,\npercuoti i tuoi nemici\nper aumentare progressivamente la quantità di danni inflitti\ncon ogni colpo a segno",
-        "koKR": "곰인간 형태일 때\n적을 강하게 후려쳐\n공격을 적중시킬 때마다\n점점 더 큰 피해를 줍니다.",
-        "plPL": "Miażdż wrogów pod postacią niedźwiedziołaka,\nz każdym kolejnym ciosem zadając\ncoraz większe obrażenia",
-        "esMX": "Cuando estás en forma de hombre oso,\nvapuleas a tus enemigos\npara infligir más daño adicional\ncon cada golpe",
-        "jaJP": "人熊形態で敵に攻撃を\n加えると、攻撃が命中するたびに\n攻撃力が増加していく",
-        "ptBR": "na forma de urso,\nesmague os inimigos\npara aumentar o dano adicional\ncom acertos sucessivos",
-        "ruRU": "Наносит врагам дополнительный урон,\nувеличивающийся с каждым ударом.\nТребуется облик медведя.",
-        "zhCN": "在熊人形态下\n槌击敌人\n后续攻击造成\n越来越高的伤害"
+        "enUS": "When in werebear form, maul your enemies\nfor increasing extra damage and savagery\nwith each successive hit.",
```
Delegus, line 30299: This is worded jankily. "When in werebear form, maul your enemies `for increasing extra damage and savagery with each successive hit.`" the word `extra` implies a global modifier such as with the charm property, or the amazons skill, if that is intended it needs to not be a double positive reinforced by descripive text and then reiterated on by additional extra words saying you get things for hitting things.

data/local/lng/strings/skills.json
```diff
@@ -34665,19 +34665,19 @@
...
-        "frFR": "Vol de vie : %+d-%d%%",
-        "itIT": "Furto di vita: %+d-%d%%",
-        "koKR": "생명력 훔치기: %+d-%d%%",
-        "plPL": "Kradzież zdrowia: %+d-%d%%",
-        "esMX": "Robo de vida: %+d-%d%%",
-        "jaJP": "ライフ吸収: %+d - %d%%",
-        "ptBR": "Roubo de pontos de vida: %+d-%d%%",
-        "ruRU": "Похищение здоровья: %+d–%d%%",
-        "zhCN": "生命偷取: %+d-%d%%"
+        "enUS": "Life Steal: %+d%% to %+d%%",
```
Delegus: The #-# is intentional to indicate a range based on charge count, putting # to # implies that it's always a range of the first number to the second number, I see this change in multiple spots.

Delegus (commented):
The linting is fine, just some things I have noted and have concerns about.

data/global/excel/base/skilldesc.txt
strangemolars: I'll only be able to change stuff on the PR monday, so I guess I'll just try to clear everything so I know what to do when I'm able.

From what I can understand, the logic you would prefer for the tooltip display would be having every static value be displayed at the top (on the right of the file) and whatever scales with soft levels (as whatever is dependent on blvl do not display properly, as I figured with the new charge effects) on the bottom (on the left of the file).

This one I changed precisely because we had several things going on at the same time. If you check current druid, you will find: 1) stuff where mana cost scale but its 'on the right' (on the top), aka, fire claw, fissure, arctic blast, cyclone armor, twister, tornado. 2) stuff where mana cost don't scale and it is on top; most summons but ravens and solar, volcano, firestorm, hurricane, armageddon and physical ss skills, rabies and 3) stuff where the mana cost scales and its on the left;  toxic fangs, moulten boulder, ravens and solar creeper only.

That was a big spread of things, and the 1) skills threw a wrench on any consistency I could use to gauge what was the preferred logic for this. Similar inconsistencies exist on the other classes as well. So I just went with displaying them at the top since that seemed to be the most common occurence. 

I can for sure change them to the preferred behavior when I get the chance monday.

data/local/lng/strings/skills.json
strangemolars: This one there was a couple reasons I chose to rework the wording for them.

First one is that I never liked the use of '-' to display the range for Feral Rage and Maul. Its something that always were on those skills, but I never liked. The reason mostly because '-' is used mostly (I think only?) on damage ranges, which is something that happens 'at the same time'. When you deal damage, you deal any number in that range, and on the next time you deal damage again, you will re-roll the range and you can have again any number inbetween that range. On Feral Rage and Maul, that is not the case. Everytime you use the skill, you accrue value on it, and you start at the manimal value, and you build (up) to the max. Changing the word is not something that fully explains what is going on, since both '-' and 'to' both are semantically the same thing, but the simple fact of writing it differently would denote a difference in function.

Second is that I didnt like how '-' displayed with the new Feral Rage functions. You would have the physical resist debuff, for example, be "-2 - -20", and I felt "-2 to -20" just looked better.

data/global/excel/base/itemstatcost.txt
strangemolars: Currently (and I think 'always' as I'm more and more positive that "extra charges" is not something I want at all to be an item mod) the max -monster armor per hit feral rage is achieving is -540 at a woping lvl 100 skill - and I dont think theres anything else on the game that reach anywhere close to those values. 

When the sendbit was 8 the stat was capping at -128. So I changed it to 12 and it started to work how I wanted, so I didnt revert to other numbers. Which number do you think would be best to leave at considering that? 9? 10?

data/global/excel/base/skills.txt
strangemolars: From what I gauged checking the files when I was thinking what to do with Toxic Fangs, the 'rabies' auratagetstate only remains on enemies as long as the Rabies effect itself remained on enemies. The state doesnt have a duration set by itself, so as soon as anything made Rabies drop on the enemies, the state will no longer exist. Toxic Fangs meanwhile would not apply any state to the enemies.

That means that before, when Toxic Fangs didnt carried this aurastate, it also made said state stop existing as soon as you applied it to any mob that had Rabies before, as the 2 sources of poison wouldnt stack and only Fangs would be on the mob. You would be cutting the -20 poison resist completely on the mob you used Fangs anyway.

I made so Fangs carry the same aurastate, meaning that now when you Fangs a mob that has Rabies (usually happens, you apply Rabies to a pack, then go to Fang a stronger target you want dead quicker). 

This mean that the 'rabies' auratagetstate will drop in .4 secons on the target you Fanged, but the idea is that you keep using Fangs and re-applying this state to thes single target, as you want it to die faster. It drops fast, but that was already the case before, you would remove the auratagetstate anyways from the mob you Fanged.

The other mobs that are being affect by Rabies meanwhile shouldn't be affected at all, right?

data/global/excel/base/skilldesc.txt
Delegus: Hmm, then one of my past changes must have been knocked out somewhere or I didn't adjust Druid during my summon rework as I thought. If the variance is that prolific then I'm not motivated enough to pursue this or make the alterations personally and can just shrug.

data/global/excel/base/skills.txt
Delegus: None of this is relevant to the question I asked. 

Take Tainted for example, state conflicts kill them if you equip or use Resist Fire while they are summoned, therefore, does the main rabies missile, aka invisible spreader, become destroyed if you rabies and then fangs the same mob? It's not entirely killing the idea as you would usually only do that on a boss, but if you hit a rare pack and want to fangs the rare while rabies kills the trash and rabies missile dies the poison it produces dies with it on every target being affected.

 If you did not observe this behavior then it is is a nonissue.

data/global/excel/base/itemstatcost.txt
Delegus: 2^9 is 511 so to acheive your goal and not inflate the value 10 should be sufficient.

data/global/excel/base/skills.txt
strangemolars: What makes rabies spread is the missile 'rabiesplague'. The targetstate 'rabies' is just the state that applies -20 poison resist to the enemies that are affected by said missile poison, its not the state that spreads the plague. If you Rabies a mob, and start Fanging another, the Rabies just keeps spreading. It even can re-infect a mob you Fanged and didnt die with Rabies again once Fangs poison drop.

data/local/lng/strings/skills.json
Delegus: Perhaps Life Steal Range: # to #

data/global/excel/base/skills.txt
Delegus: The rabiesplague is the originator of the state and the state dies with it.

strangemolars: I'm not understanding how Toxic Fangs also applying the same state would kill the targetstate on mobs you are not hitting with it and even less how it would kill the rabiesplague spawner in any way it wouldn't already do before.

Delegus: Via state conflict, if you use the same state for more than one source it tries to remove one, hence why all auras all use different states, sometimes two different if they affect allies and enemies.

data/global/excel/base/itemstatcost.txt
strangemolars: changed to 10.

data/global/excel/base/skilldesc.txt
strangemolars: Updated all skills to follow the preferred logic.

data/global/excel/base/skills.txt
strangemolars: Updated to a new state just in case, on a new state group with the rabies one so they dont stack. Tested behavior and it seems to be working as it always did.

data/local/lng/strings/skills.json
strangemolars: Changed to a per charge display. With no absurd charge#s and a max charges display, no need for the ranges IMO.

Delegus (approved):
Lints fine, LGTM

PR #1312, opened 2026-08-27
small fix for new feral rage and maul

strangemolars:
(Pre-ship fix for pull 1307 that I messed up last second)

When I was about to ship the skills, I made a last change to the formulas by changing the references to the skill itself to just .lvl

I forgot I had already gone through the frustration of understanding these skills in particular understand the shortcut ".lvl" as CHARGE LEVEL and not SKILL LEVEL on its calculations, and that broke the scaling of the skills.

So I fixed them back, but using skill('xxx'.lvl) breaks the next level tooltip display, so I made 2 mock columns to fix that as well. What we brazillians like to call "gambiarra".

Tried to future-proof people re-breaking these skills by adding some OBS to the description columns, as these skills just don't respect usual D2 euristics. If its too much, we can remove it.


CollinHerber (approved):
LGTM

PR #1317, opened 2026-09-01
feature/bugfix-sunder-recipe-black-cleft

The-Honorable:
# Bugs

- fix cubemain description for black cleft sunder recipe


The-Honorable:
also fixed in #1321

PR #1328, opened 2026-09-17
Bug Fix & Skele Nerf w/ Forb Nerf & Jewel Ner-Buff

Delegus:
### Necromancer
- Raise Skeleton
  - Adjusted health to sane values, updated tooltips formula to reflect the true health the summons have.
- Raise Skeletal Mage
  - Adjusted health to sane values, updated tooltips formula to reflect the true health the summons have.

>[!Note]
>This little rabbit hole + some advice from Ben  and a post from celestialray has shown we have a severe over-stat problem with all minions attack rating and defense we will address in a later pull request and patch.

### Sorceress
- Frozen Orb
  - Reduced explosion radius of individual bolts from 5 to 3

>[!Note]
> These bolts were way overperforming in big groups, if this doesn't tone it down enough then we will start adjusting the curves down.

## Menu
- Removed obsolete mainmenupanelHD.json causing positioning problems for D2RLoader, verified no issues non-loader launch visually

## Items
- Unique Colossal Jewels
  - Removed the CTC skills as all but 1 were broken or unwanted and the 1 desirable skill was the only jewel anyone wanted.
  - Added sunder to these jewels to match their corresponding elements as a source of sunder to free up a charm slot or give to your mercenary so they can do damage to immune enemies too.
  - Physical Damage jewel was reduced to 75-100% EWD and 20-40 Min/Max from 100-150 and 30-50
  - Increased the gold find %

>[!Note]
>Probably going to revisit these in the future when I have more time/motivation because I want to create a minion related one with an aura that only affects friendly monsters and buffs them in some manner, or gives an oSkill similar to the warlock death mark


The-Honorable (approved):
lgtm

